PR #465, opened 2017-01-04
Restrict inline image height to the text height

thewilkybarkid:
Images inside text are a problem. Currently these images are just included as is, leading to:

screen shot 2017-01-04 at 15 08 37

This restricts them to the height of the text:

screen shot 2017-01-04 at 15 08 46

This doesn't attempt to resize images inside other elements (eg `<td>`, `<li>`), as they're likely not text-height (or we have no way of determining it).

### DIFF
--- a/assets/sass/_mixins.scss
+++ b/assets/sass/_mixins.scss
@@ -264,3 +264,7 @@
   }
 }
 
+@mixin inline-image {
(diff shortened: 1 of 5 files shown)


thewilkybarkid:
@davidcmoulton Closing this due to the fork problem, but probably should still be done.

PR #524, opened 2017-03-08
Don't nest styles in article sections

thewilkybarkid:
We're soon to allow buttons appearing in article sections. Currently the colour of the button is overwritten due to the nested of styles inside article sections. I don't understand why this is all article-section specific, so have tried to move it to the top level to avoid such specificity problems.

There might be some subtle changes elsewhere as a result, I've tried to look through and fix what I can.


thewilkybarkid:
Closing this as it's out of date.

PR #693, opened 2017-06-05
ELPP-2791 Only use MutationObserver if it's available

thewilkybarkid:
Decided not to block AssetNavigation, means that any captions containing MathML will probably appear un-MathJaxed. Given that we haven't addressed non-JS MathML anyhow, let's just live with it.

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -5,7 +5,7 @@ require('./libs/polyfills');
 // Base level of feature support needed for the js loaded in this file.
 // Consider AJAXing in the rest if the test passes.
 if (window.localStorage && document.querySelector &&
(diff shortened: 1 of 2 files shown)


assets/js/main.js
```diff
@@ -5,7 +5,7 @@ require('./libs/polyfills');
 // Base level of feature support needed for the js loaded in this file.
 // Consider AJAXing in the rest if the test passes.
 if (window.localStorage && document.querySelector &&
-       window.addEventListener && !!(document.createElement('div')).classList) {
+       window.addEventListener && !!(document.createElement('div')).dataset) {
```
thewilkybarkid: The only case I can see where something supports `dataset` but not `classList` is Opera 11.10 to 11.50. 🤷‍♂️

PR #697, opened 2017-06-05
ELPP-2789 Link First Author et al to details section (narrow)

davidcmoulton:
- Make `authorLine` content populate a real DOM element, for use on narrow view
- Enable linking of the author line, and account for flag to indicate whether "et al." should also be shown.

@thewilkybarkid: this will require changes to `patterns-php`, let me know if you see anything in here that may cause problems with that.

### DIFF
--- a/source/_patterns/02-organisms/content-headers/content-header.mustache
+++ b/source/_patterns/02-organisms/content-headers/content-header.mustache
@@ -127,12 +127,6 @@
     </div>
   {{/authors}}
 
-  {{^authors}}
-    {{#authorLine}}
-      <div class="content-header__authors content-header__authors--line">{{{authorLine}}}</div>
-    {{/authorLine}}
-  {{/authors}}
-
   {{#button}}
     <div class="content-header__cta">
       {{>atoms-button}}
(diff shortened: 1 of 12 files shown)


thewilkybarkid:
The author line on larger screen for the read more header is hidden.

source/_patterns/02-organisms/content-headers/content-header.mustache
```diff
@@ -127,11 +127,18 @@
...
-  {{/authors}}
+  {{! authorLine used to link to main article author section when narrow }}
+  {{#authorLine}}
+    <!-- aria-hidden because this is just a short summary for narrow viewports, the full author list should be accessible -->
+    <div class="content-header__authors content-header__authors--line" aria-hidden="true">
+      {{#url}}
+        <a href="{{url}}" class="content-header__author_link">{{{text}}}{{#hasEtAl}}<span class="content-header__author_link_highlight">  et al.</span>{{/hasEtAl}}</a>
+      {{/url}}
+      {{^url}}
+        {{{text}}}
```
thewilkybarkid: This needs to add et al

PR #699, opened 2017-06-16
ELPP 2864 author list when narrow

davidcmoulton:
*WIP as awaiting Product approval.*

This update:

- implements new behaviour for the author list within an article's main content header;
- adds/amends tests for the new behaviour;
- adds some missing tests, and
- tidies the code a little.

### DIFF
--- a/assets/js/components/ContentHeader.js
+++ b/assets/js/components/ContentHeader.js
@@ -1,4 +1,5 @@
 'use strict';
+const utils = require('../libs/elife-utils')();
 
 module.exports = class ContentHeader {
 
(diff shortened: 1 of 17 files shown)


thewilkybarkid:
First author seems to lose their comma:

screen shot 2017-06-19 at 08 39 24

thewilkybarkid:
The popup triggers on larger screens, but also follows the link.

davidcmoulton:
@thewilkybarkid Comma fixed. I'm going to ignore the errant behaviour in PatternLab, as the behaviour is correct when not within PL's environment.

PR #700, opened 2017-06-26
Restrict universal selector to top level

thewilkybarkid:
Stops it bleeding into MathJax etc (switch to CommonHTML). May have side effects.


davidcmoulton:
To get a sense of the possible side effects, have you checked any instances of the `captioned-asset` pattern where there's nested HTML injected into `caption-text`?

thewilkybarkid:
We currently can have a `<p>`, `<div class="table-wrap">` or `<div class="math-block">`, so I reckon it's ok.

davidcmoulton:
Alright, will merge, but we should keep an eye out there for any unanticipated craziness.

PR #703, opened 2017-06-29
ELPP-2624 About profiles grid

thewilkybarkid:
(Depends on #702).


assets/sass/patterns/molecules/popup.scss
```diff
@@ -6,7 +6,8 @@
 
 .popup {
   position: absolute;
-  width: 340px;
+  min-width: 340px;
+  max-width: #{$bkpt-site--medium - 10 - 10}px; // threshold - left - right
```
thewilkybarkid: The text in the about profile popups can be quite long, so this makes them bigger. This also changes the other popups, which might not be desirable.

assets/js/components/Popup.js
```diff
@@ -3,8 +3,10 @@ const utils = require('../libs/elife-utils')();
 module.exports = class Popup {
 
   constructor($elm, _window = window, doc = document) {
-    if (!$elm.hash || $elm.host !== _window.location.host || !$elm.hash.match(/^#[a-z]/i)) {
-      return;
+    if (!$elm.dataset.popupSelf) {
```
thewilkybarkid: Introducing this feels a bit like a hack, but the element isn't an `<a>` and it doesn't have an ID.

assets/sass/patterns/molecules/about-profile.scss
```diff
@@ -19,18 +41,22 @@
...
-
-  .about-profile__picture {
+  @media only all and (min-width: 700px) {
     float: right;
     width: 250px;
     @include margin(24, "left");
   }
 
+  @media only all and (min-width: #{$bkpt-site--medium}px) {
+    .about-profiles__item .about-profile--popup & {
```
thewilkybarkid: Not happy with this hack.

assets/sass/patterns/molecules/about-profile.scss
```diff
@@ -19,18 +41,22 @@
...
-}
 
-.about-profile__image {
-  width: 100%;
-}
-
-@media only all and (min-width: 700px) {
-
-  .about-profile__picture {
+  @media only all and (min-width: 700px) {
```
davidcmoulton: `min-width: #{$bkpt-site--medium}px`?

giorgiosironi:
Got to the point in which both the popup behavior (large viewport) and collapsable caption (small viewport) work without interfering with each other. Now onto trying to introduce dynamically a `<a>` around the popup, which is difficult.

giorgiosironi:
Also, the popup tests have some problem with the media query failing.

assets/sass/patterns/molecules/about-profile.scss
giorgiosironi: As far as I understand this hides the element `.about-profile--popup` content embedded in the page, as long as we are in the "large page" scenario. Then we clone that element when the popup needs to be opened, putting the clone in a special place in the page. But then the content ends up being present twice in the page, which is strange.

giorgiosironi:
I am expanding the test coverage for Popups, because they have lots of behavior (only one open at a time apparently, events that propagate to the document level) but they had 0 tests.

assets/js/components/ToggleableCaption.js
```diff
@@ -17,6 +18,13 @@ module.exports = class ToggleableCaption {
       return;
     }
 
+    if (this.thresholdWidth != null) {
```
davidcmoulton: Use either `if (this.thresholdWidth)` if any truthy value will do, or if you're specifically wanting to check for the absence of `null`, use `if (this.thresholdWidth !== null)`

assets/sass/_normalize-overrides.scss
```diff
@@ -145,6 +145,10 @@ dd {
   margin-left: 0;
 }
 
+dd + dd {
+  @include margin(-24, "top");
```
davidcmoulton: Negative margins can result in unanticipated overlays, is there a better way of achieving the desired effect?

assets/js/components/ToggleableCaption.js
```diff
@@ -17,6 +18,13 @@ module.exports = class ToggleableCaption {
...
+      const conditionMediaQuery = `(max-width: ${this.thresholdWidth}px)`;
+      if (!this.window.matchMedia(conditionMediaQuery).matches) {
```
giorgiosironi: we should listen for changes in viewport size so that we can check dynamically whether this media query is still/not satisfied. Same goes for Popup.js

assets/js/components/Popup.js
```diff
@@ -3,11 +3,24 @@ const utils = require('../libs/elife-utils')();
...
   constructor($elm, _window = window, doc = document) {
-    if (!$elm.hash || $elm.host !== _window.location.host || !$elm.hash.match(/^#[a-z]/i)) {
-      return;
+    var $link = $elm;
+    if ($elm.dataset.popupWrapper) {
+      const wrapper = utils.buildElement('a');
+      $elm.parentNode.insertBefore(wrapper, $elm);
+      wrapper.appendChild($elm);
+
+      wrapper.href = '#some-value'; // can't be removed
```
giorgiosironi: this should work by putting just `#` in it?

giorgiosironi: yes, by `.preventDefault()` it

assets/js/components/Popup.js
```diff
@@ -3,11 +3,24 @@ const utils = require('../libs/elife-utils')();
 module.exports = class Popup {
 
   constructor($elm, _window = window, doc = document) {
-    if (!$elm.hash || $elm.host !== _window.location.host || !$elm.hash.match(/^#[a-z]/i)) {
-      return;
+    var $link = $elm;
+    if ($elm.dataset.popupWrapper) {
+      const wrapper = utils.buildElement('a');
```
giorgiosironi: add a class so that the "everything inside this is blue" is disabled

giorgiosironi: pseudo classes included

assets/sass/_normalize-overrides.scss
```diff
@@ -145,6 +145,10 @@ dd {
   margin-left: 0;
 }
 
+dd + dd {
```
davidcmoulton: Prefer `.about-profile__profile dd + dd`; move to appropriate `scss` file.

assets/sass/patterns/molecules/about-profile.scss
```diff
@@ -1,16 +1,55 @@
...
+  @media only all and (min-width: #{$bkpt-site--medium}px) {
+
+    .about-profile--popup & {
+      display: none;
+
+      .popup & {
+        display: block;
+        margin-top: 0;
+
+        & p,
```
davidcmoulton: There's some pretty funky `&` use going on here. Bit confusing, should be simplified.

source/_patterns/01-molecules/components/about-profile.mustache
```diff
@@ -26,9 +26,12 @@
 
   {{#profile}}
 
-    <div class="about-profile__profile">
 
+    <div class="about-profile__profile " data-behaviour="ToggleableCaption" data-threshold-width="730">
```
davidcmoulton: Insert a comment indicating that 730 is a site breakpoint `$bkpt-site--medium`.

thewilkybarkid:
Think this requires some refactoring, but it works and meets the spec so merging for now.

PR #704, opened 2017-06-30
[WIP] Teaser variant with footer reference

nlisgo:


thewilkybarkid:
Might be simpler to just include a rendered list inside `content` as this only applies in one place?

nlisgo:
@davidcmoulton if you agree with this approach I could do that.

davidcmoulton:
@nlisgo I'm happy to proceed with the suggestion proposed by @thewilkybarkid for now. If we have something additional similar requirement we can revisit later.

PR #709, opened 2017-07-05
ELPP-2861 Support list being added to teaser content

nlisgo:


assets/sass/patterns/atoms/list.scss
```diff
@@ -18,3 +18,7 @@ $prefixes: ("alpha-lower": "lower-latin", "alpha-upper": "upper-latin", "bullet"
     }
   }
 }
+
+.list--teaser {
+  @include padding($blg-space-small-in-px, "top");
```
thewilkybarkid: Is this meant to be separating the list from the text? Or are there more requirements not yet fulfilled? If it's just this, can the text just be wrapped in a `<p>` instead?

nlisgo: I have no objection to that approach but @davidcmoulton didn't seem keen on the idea. I can confirm that wrapping the text in a `<p>` element gives the text and list the separation that it needs.

davidcmoulton: @nlisgo I'm not against the idea, I was saying that semantically the text didn't need to be wrapped any further than it was already. Sorry for the misunderstanding.

If wrapping the text in `<p>` gives what you need for vertical spacing, so then the `list--bullet` variant can just be injected as content then that's fine. That reduces the code changes to `pattern-library` considerably.

Will the text always be wrapped in a `<p>`, or only if there's a list following? If always (preferable, I think), then check if this affects the vertical spacing of teasers that don't have a list in the content (i.e. does the `<p>`add extra, unwanted space between these teasers?).

nlisgo: @davidcmoulton there is already a `margin-bottom` of 23px (1.4375rem) on `.teaser`. If I wrap the text in a `<p>` and don't add a list then the only visible difference is caused by the `margin-botton` on the `<p>` of 24px (1.5rem).

I was only considering adding the `<p>` tag for annual-reports and they will all have a list underneath the impact statement now.

source/_patterns/00-atoms/components/list~teaser-content.json
```diff
@@ -0,0 +1,9 @@
+{
+  "isOrdered": false,
+  "prefix": "bullet",
+  "classes": "list--teaser",
+  "items": [
+    "item 1",
```
davidcmoulton: As the pattern variant addresses items containing links, please include a link in one of these item examples.

nlisgo:
@davidcmoulton I have made the requested changes.

PR #711, opened 2017-07-07
PI-192 Add sequence class

thewilkybarkid:
Not working in Edge/IE at least, even though it should (http://caniuse.com/#search=word-break).


thewilkybarkid:
`display: inline-block` causes it to work in Edge, but obviously that's not right.

thewilkybarkid:
Suspect this from [`word-wrap`](https://msdn.microsoft.com/en-us/library/ms531186(v=vs.85).aspx) applies to `word-break`:

> This property applies to elements that have layout. An element has layout when it is absolutely positioned, is a block element, or is an inline element with a specified height or width.

thewilkybarkid:
Just spotted same happens on Firefox. :disappointed:

thewilkybarkid:
`word-wrap: break-word` works for both, but starts it on a new line rather. Might be the best we can do.

thewilkybarkid:
The spec says [all elements](https://drafts.csswg.org/css-text-3/#word-break-property), so looks to be bugs.

thewilkybarkid:
https://bugzilla.mozilla.org/show_bug.cgi?id=1283964

PR #712, opened 2017-07-18
Tidy up article toggle

giorgiosironi:
Started as equal widths but turned into a tabbed menu. There may be some more variables that can be referenced in the CSS in place of hardcoded values, or `em` values, or Sass mixins.

Notable points:
- `ul` now expands to 100% width to provide bottom border
- `li` are still floated and align to the left with some margins
- the active `li` has a white border moved to delete out the one from the `ul`
` `li` are still hovered in the non-active version to show the clickable area with borders
- `a` elements now expand to the whole of their `li` container so that the largest possible area can be clicked
- currently active `a` is bold, hovered one is black but not bold (determined by hovering over the larger `li`, not itself)

### DIFF
--- a/assets/sass/patterns/molecules/view-selector.scss
+++ b/assets/sass/patterns/molecules/view-selector.scss
@@ -130,11 +130,13 @@
   .view-selector__list-item--article {
     border-right: none;
     @include border-radius-simple(4, 0, 0, 4);
+    width: 170px;
   }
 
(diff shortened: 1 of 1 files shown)


davidcmoulton:
The requirements for this have changed: see today's update to ticket ELPP-2799.

giorgiosironi:
Well it was a good learning experience :laughing:

PR #720, opened 2017-07-27
ELPP-3013 Inline image height

thewilkybarkid:
Reopened #465.


assets/sass/_mixins.scss
```diff
@@ -368,3 +368,8 @@
   @include font-size(14);
   line-height: #{(20/14)};
 }
+
+@mixin inline-image {
+  margin-bottom: -0.225em;
+  height: 1em;
```
thewilkybarkid: I think this is the best we can do. It's not 1 line height, but we need to be relative. Seems to work in the cases I've tested.

PR #723, opened 2017-07-28
ELPP-3010 Fix Fragment handling

thewilkybarkid:
Two problems:

1. Lazily-loading the JS means that `DOMContentLoaded` might have already fired.
2. A race-condition between `FragmentHandler` and `ArticleSection` sees sections closing after the fragment is handled.


assets/js/components/ArticleSection.js
```diff
@@ -74,7 +84,7 @@ module.exports = class ArticleSection {
     this.$headerLink.classList.remove('article-section__header_link--closed');
     this.$elm.classList.remove('article-section--collapsed');
     this.$body.classList.remove('visuallyhidden');
-    if (!!this.window.MathJax) {
+    if (!!this.window.MathJax && !!this.window.MathJax.Hub) {
```
thewilkybarkid: Unrelated, but was causing sporadic errors.

assets/js/components/FragmentHandler.js
```diff
@@ -15,54 +15,39 @@ module.exports = class FragmentHandler {
     this.window = _window;
     this.doc = doc;
     this.isSingleton = true;
-    this.window.addEventListener('DOMContentLoaded', this.handleDomLoad.bind(this));
+    this.window.addEventListener('readystatechange', this.onReadyStateChange.bind(this));
+    this.onReadyStateChange({});
```
thewilkybarkid: We can't use `DOMContentLoaded` as it might have fired before this is executed.

assets/js/components/FragmentHandler.js
```diff
@@ -15,54 +15,39 @@ module.exports = class FragmentHandler {
...
    * @param {String} idToFind The id to search for
    * @param {HTMLDocument} doc
-   * @param {Function} areElementsNested Handles determination of nesting
    * @returns {String} The id of the collapsed section containing idToFind, or null
    */
-  getIdOfCollapsedSection(idToFind, doc, areElementsNested) {
+  getIdOfCollapsedSection(idToFind, doc) {
     let collapsedSections = doc.querySelectorAll('.article-section--collapsed');
-    if (!collapsedSections) {
+    if (!collapsedSections.length) {
```
thewilkybarkid: Unrelated, but was never working.

assets/js/components/FragmentHandler.js
```diff
@@ -15,54 +15,39 @@ module.exports = class FragmentHandler {
...
       return null;
     }
 
     let $collapsedSectionContainingFrag;
     [].forEach.call(collapsedSections, ($collapsedSection) => {
       if (!$collapsedSectionContainingFrag) {
-        if (this.isIdOfOrWithinSection(idToFind, $collapsedSection, doc, areElementsNested)) {
```
giorgiosironi: perhaps we should move in the future to tell all sections `this fragment has been selected` rather than telling a single one `open`. That would move keep this logic inside `ArticleSection`, but it's out of scope for the bugfix

test/articlesection.spec.js
```diff
@@ -40,6 +40,60 @@ describe('An Article Section (collapsible)', function () {
     expect(articleSection.$body.classList.contains('visuallyhidden')).to.be.true;
   });
 
+  it('is initially opened if initial state is closed but the fragment is the section', function () {
+    $elm.dataset.initialState = 'closed';
+    let windowMock = {
```
giorgiosironi: can this mock be extracted at the test case level, to be easily created in different tests (e.g. `let windowMock = withHash('#introduction')`)?

test/fragmenthandler.spec.js
```diff
@@ -30,42 +30,6 @@ describe('A FragmentHandler Component', function () {
...
-      let docMock;
-
-      beforeEach(function () {
-        docMock = {
-          querySelector: function () {},
-          querySelectorAll: function () {}
-        };
-      });
-
-      it('returns true if the id tested matches the section id', function () {
```
giorgiosironi: these tests should go into the `elife-utils` unit tests?

thewilkybarkid: 👍

assets/js/components/FragmentHandler.js
thewilkybarkid: I like this idea, a generic 'node selected' event (maybe there's a better word than selected?). This can remove the `expandsection` event below, since the `ArticleSection` component can listen to it and open as needed.

But yeah, not part of this bug fix.

PR #726, opened 2017-08-01
ELPP-2844 Remove bold from content header impact statements

thewilkybarkid:


assets/sass/_typographic-hierarchy.scss
```diff
@@ -180,7 +180,7 @@
 @mixin subtitle-typeg() {
   font-family: $font-secondary;
   @include font-size(16);
-  font-weight: 700;
+  font-weight: 500;
```
thewilkybarkid: Making the change here as it's the only usage of `subtitle-typeg` currently. Change to an override if necessary later.

Also leaving `font-weight` in for consistency (eg matches `institute-typeg`).

PR #727, opened 2017-08-01
ELPP-2964 Allow captioned assets to be inlined

thewilkybarkid:


thewilkybarkid:
Might be better to have a wrapper rather than a variant.

thewilkybarkid:
Hopefully caught all the things that need to clear.

PR #729, opened 2017-08-02
[WIP] Use display-density responsive images

thewilkybarkid:
We try and use `srcset` with width descriptors, but rarely provide the required `sizes` attribute, or one that's correct.

The pattern (usually) can't know what `sizes` should be, since it depends where the pattern is used. This would leave the user to try and work it out, which is _very_ messy. (There's a handful of places where we correctly use `sizes`, but only because they're fixed-width images.)

When using the patterns in production, we're just trying to show 1x and 2x images. So, let's simplify things and use display-density descriptors instead.


thewilkybarkid:
We might want to make it 1x and any number of other sizes. We don't always have an image that's 2x as big (but would be useful to have a 1.5x instead), plus would allow for 3x etc.

davidcmoulton:
Closing in favour of https://github.com/elifesciences/pattern-library/pull/770

PR #730, opened 2017-08-02
ELPP-2831 Increase space above meta info

giorgiosironi:
We obtain this by using two margins around secondary info and footer. If the teaser body (impact statement) in between the two is missing, the two margins collapse resulting in still only 12px of distance

### DIFF
--- a/assets/sass/patterns/molecules/teaser.scss
+++ b/assets/sass/patterns/molecules/teaser.scss
@@ -87,7 +87,6 @@
 
 .teaser__secondary_info {
   @include listing-main-author-typeg();
-  @include padding(0, "bottom");
   @include margin(12, "bottom");
 }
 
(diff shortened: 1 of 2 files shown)


assets/sass/patterns/molecules/teaser.scss
```diff
@@ -87,7 +87,8 @@
 
 .teaser__secondary_info {
   @include listing-main-author-typeg();
-  @include padding(12, "bottom");
+  @include padding(0, "bottom");
```
thewilkybarkid: Not needed.

giorgiosironi: in the sense that it can just be removed and it will be `0` by default?

thewilkybarkid: Yep.

PR #731, opened 2017-08-03
ELPP-2651 Add inline profile

thewilkybarkid:


thewilkybarkid:
@davidcmoulton We missed adding the YAML file!

PR #733, opened 2017-08-03
ELPP-2820 ELPP-2412 Teaser format

thewilkybarkid:
Current the teaser pattern tries to say whether the article is in PoA or VoR, but these don't line up with 'PDF', 'HTML' or 'HTML + PDF' which we actually need to display.


source/_patterns/01-molecules/teasers/teaser.yaml
```diff
@@ -89,13 +89,14 @@ schema:
       properties:
         meta:
           $ref: ../../01-molecules/components/meta.yaml#/schema
-        publishState:
+        formats:
           type: object
+          minProperties: 1
           properties:
-            vor:
+            html:
```
giorgiosironi: new format is more extensible, as new keys can be added without changing the existing ones

giorgiosironi: with respect to an array `['html', 'pdf']`, makes clear that the input order doesn't count as it's decided in the template

PR #735, opened 2017-08-03
ELPP-3018 Add no-image placeholder graphic to archive listings

giorgiosironi:
- Added images from ELPP-3018 ticket
- One example of listing now hasn't got an image
- Correcting non-BEM class name
- Adding default image as background

To be done:
- the transparent/white is not applied on :hover, and the text becomes almost as black as the background making this a problem
- `image-set` may be used in the CSS to serve the 2x image


thewilkybarkid:
Don't worry about 2x, I don't think there's any browser that doesn't support SVG but does support `image-set`.

Wonder if this should always be the case, so if an image fails, for example, it remains the same.

giorgiosironi:
I initially tried to put it always there, but there are other elements (maybe it was the semi-transparent overlay) that need to coordinate their z-indexes so that this image stays below everything.

thewilkybarkid:
I got it working when the image failed to load/was loading, but the background on `.block-link__picture` still takes effect, making it darker than normal. Compromise is to make the background black in these cases.

assets/sass/patterns/atoms/block-link.scss
```diff
@@ -116,9 +120,9 @@
     bottom: 0;
     left: 0;
     z-index: -1;
-    background-image: linear-gradient(to top, rgba(0, 0, 0, 0.5), rgba(0, 0, 0, 0.5));
+    background-color: rgba(0, 0, 0, 0.5);
     .block-link:hover & {
-      background-image: linear-gradient(to top, rgba(255, 255, 255, 0.5), rgba(255, 255, 255, 0.5));
+      background-color: rgba(0, 0, 0, 0.8);
```
davidcmoulton: Use sass variable `$color-overlay` for this, so: `background-color: $color-overlay;`

giorgiosironi:
Changes made. We could also extract another variable for https://github.com/elifesciences/pattern-library/pull/735/files#diff-73f529139aa98260151ab06c654c1aedR132

davidcmoulton:
@giorgiosironi I don't think that's a candidate for extraction yet as that value's only used in this context once. Can re-evaluate if it's used anywhere similar in the future.

PR #738, opened 2017-08-07
Providing static website demo of every pull request, updated in place on each new commit

giorgiosironi:
1. Go to `All checks have passed`
2. Click on `Show all checks`
3. Click the `Details` link of `continuous-integration/jenkins/pr-demo`
4. Enjoy the demo with shareable URL to send to product

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,13 +15,14 @@ elifePipeline {
 
             elifePullRequestOnly { prNumber ->
                 stage 'Deploying to a public URL', { 
+                    def url = "https://s3.amazonaws.com/ci-pattern-library/${prNumber}/index.html"
+                    elifeGithubCommitStatus commit, 'pending', 'continuous-integration/jenkins/pr-demo', 'Static website is being built', url
                     builderCmd 'pattern-library--ci', './archive-public.sh', '/srv/pattern-library' 
                     sh 'scp -o StrictHostKeyChecking=no elife@ci--ui-patterns.elifesciences.org:/srv/pattern-library/public.tar.gz .'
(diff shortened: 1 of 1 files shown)


elife-alfred-user:
I'm elife-alfred-user and I want to post a comment here (manually through the Github UI).

elife-alfred-user:
You can see this pattern-library version at https://s3.amazonaws.com/ci-pattern-library/738/index.html

giorgiosironi:
First thought: store state on the filesystem as we do for events. Won't work very well as libraries are build on slave nodes rather than on the master.

giorgiosironi:
Looking for an existing comment with a similar text is slower (one more API call) but reliable, it is not flexible enough to be reused for other actions but a good-enough solution for this case.

thewilkybarkid:
Could we do what we had with Platform.sh, so it's a separate build? There wouldn't be that problem then.

Think it would be useful to separate code checks and tests in the same way.

giorgiosironi:
We can have a separate and parameterized build triggered with the commit sha from here once the tests pass (or even before); but there is a dependency as the output used to populate the bucket is the current content of `pattern-library--ci` so we would have to do the same Salt highstate twice. Would also still be 1 comment for every commit, which is not what we want.

Unless we want to trigger it manually, but it's annoying to click around for every update to the code that needs to be published.

thewilkybarkid:
I mean the link to the deploy is the build's 'Details' link, rather than a comment.

giorgiosironi:
We have some code around to notify arbitrary Git commit statuses, will check if it works in this case.

giorgiosironi:
Ok, provided in the `Static website` link.

PR #739, opened 2017-08-08
ELPP-2802 Change error spacing and make it centred

thewilkybarkid:
Increases the spacing before/after error messages, but also changes them to be horizontally/vertically centred in their container. This allows the static error pages (eg https://s3.amazonaws.com/prod-elife-error-pages/4xx.html) to not have to define local CSS to do so (non-Flexbox-supporting browsers lose the vertical centring, but that's not a problem). Journal pages (eg https://elifesciences.org/foo) are unaffected, since the container there is not full height.


assets/sass/patterns/organisms/content-header.scss
```diff
@@ -1,9 +1,8 @@
 @import "../../definitions";
 
 .content-header {
+  @include wrapper(false);
```
thewilkybarkid: Can't see any side-effects.

source/_patterns/00-atoms/errors/client-error.mustache
```diff
@@ -1,8 +1,10 @@
-<div class="error">
+<div class="error-wrapper">
+
+  <div class="error">
 
     <picture>
         <source srcset="{{#assetRewrite}}{{assetsPath}}/img/errors/4xx.svg{{/assetRewrite}}" type="image/svg+xml">
-        <img srcset="{{#assetRewrite}}{{assetsPath}}/img/errors/4xx-2x.png{{/assetRewrite}} 266w" src="{{#assetRewrite}}{{assetsPath}}/img/errors/4xx.png{{/assetRewrite}}" sizes="133px"
+        <img srcset="{{#assetRewrite}}{{assetsPath}}/img/errors/4xx-2x.png{{/assetRewrite}} 86w" src="{{#assetRewrite}}{{assetsPath}}/img/errors/4xx.png{{/assetRewrite}}" sizes="43px"
```
thewilkybarkid: These changes are unrelated.

PR #746, opened 2017-08-15
ELPP-2966 related content placement

davidcmoulton:


assets/sass/patterns/organisms/listing.scss
```diff
@@ -34,6 +34,21 @@ $totalHighlightsSpacing: 6;
   .content-header--read-more {
     border: none;
   }
+
+  // When wide, hide to avoid duplication: then, the contents is also displayed in the right hand article column
+  .listing-list__item--related:first-child {
```
thewilkybarkid: This suggests to me that the variant should be on the list, rather than the item.

davidcmoulton: Would that be granular enough? Only one item in the read more list will be a related item, that item is indicated by this `--related` variant.

(It's expected to be the first item., but we could lose the `:first-child` so the related item would still be hidden if it was in a different list position.)

thewilkybarkid: I'd say either lose the `:first-child` and allow for any item being related (unlikely), or be explicit and use something like `.listing-list--has-related > :first-child`.

davidcmoulton: I'm happy to remove `:first-child`, as it's the duplication we want to avoid, and doing so will not tie the matching of this selector to the related item's list position.

PR #749, opened 2017-08-21
ELPP-2838 Improve form validation states

thewilkybarkid:


assets/js/components/Carousel.js
```diff
@@ -41,7 +41,6 @@ module.exports = class Carousel {
     this.updateWidth();
     this.updateControlPanel(this.currentSlide);
     this.togglePlay(this.$toggler);
-    this.$toggler.focus();
```
thewilkybarkid: Doesn't seem to make a difference, but this stops `autofocus` working otherwise.

giorgiosironi: we have `autofocus` on which form? On the carousel itself for changing its screen?

thewilkybarkid: After submitting the Email CTA with invalid data.

PR #750, opened 2017-08-24
ELPP-3098 Fix fragment on page load when using ToggleableCaption

thewilkybarkid:
Now works correctly on Chrome, but not yet Firefox.


thewilkybarkid:
Same problem in Edge. Seems like the fragment scroll is happening before the captions are truncated.

thewilkybarkid:
Firefox/Edge problem caused by `async`.

thewilkybarkid:
I've tried to store what the initial fragment is, and scroll to it after the JS has finished. It appeared to work in a basic PL page, but when putting it on a real page (with significantly more content) it no longer works.

thewilkybarkid:
Struggling to find a solution. `AssetNavigation` uses promises, so I've tried to scroll to the fragment _after_ they've resolved. However, that still doesn't take image loading into account, so it works out the position then it's pushed down as they load.

thewilkybarkid:
Chrome works because of scroll anchoring (introduced in Chrome 56, it doesn't work in 55 and before).

thewilkybarkid:
Going to merge this for now, since it un-breaks Chrome (and Opera).

PR #752, opened 2017-08-30
ELPP-3130 Improve form messaging

davidcmoulton:
If we're not using native browser validation, we need to indicate to AT when a form input is invalid.

Now updated to add a a third state: `info` that allows informational messages with form items.


source/_patterns/00-atoms/forms/select.yaml
```diff
@@ -55,3 +62,4 @@ schema:
     - options
   dependencies:
     label: [id]
+    messageId: [isInvalid]
```
thewilkybarkid: Other way around? `isInvalid` requires `messageId`?

source/_patterns/00-atoms/forms/select.mustache
```diff
@@ -12,14 +12,18 @@
...
+    {{/isInvalid}}
   >
     {{#options}}
       <option value="{{value}}" {{#selected}}selected{{/selected}}>{{displayValue}}</option>
     {{/options}}
   </select>
 
   {{#message}}
-    <div class="form-item__message">{{{.}}}</div>
+    <div class="form-item__message" id="{{messageId}}">{{{.}}}</div>
```
thewilkybarkid: `messageId` is optional.

source/_patterns/00-atoms/forms/select.yaml
thewilkybarkid: Could be bidirectional? `messageId` is superfluous is there's no `isInvalid`.

davidcmoulton: The intent is that if `isInvalid` is present (it can only be true if it is), then there must be a `messageId`.

thewilkybarkid: `isInvalid: [messageId]`

source/_patterns/00-atoms/forms/select.mustache
```diff
@@ -12,14 +12,16 @@
     {{#autofocus}} autofocus{{/autofocus}}
     {{#required}} required="required"{{/required}}
     {{#disabled}} disabled="disabled"{{/disabled}}
+    {{#userInputInvalid}} aria-invalid="true"{{/userInputInvalid}}
```
davidcmoulton: Keeps the invalid state indication separate from the messaging logic, so an informational message can be provided when the form is not invalid.

source/_patterns/00-atoms/forms/select.yaml
```diff
@@ -55,3 +62,6 @@ schema:
     - options
   dependencies:
     label: [id]
+    userInputInvalid: [message]
```
davidcmoulton: Lines 65-67 keeps the invalid state indication separate from the messaging logic, so an informational message can be provided when the form is not invalid.

source/_patterns/01-molecules/components/compact-form.yaml
```diff
@@ -17,13 +17,24 @@ schema:
       enum:
         - GET
         - POST
-    state:
+    variant:
       type: string
       enum:
         - valid
         - error
+        - info
```
thewilkybarkid: I think we should lose this, and leave `state`. (It's always one of valid, invalid or not yet completed.)

source/_patterns/01-molecules/components/compact-form.yaml
```diff
@@ -17,13 +17,24 @@ schema:
...
     message:
-      type: string
+      type: object
+      properties:
+        text:
```
thewilkybarkid: I think we need two things: help text + message. (Message probably only populated when an error? Probably no need to enforce that here. Could have a dependency on `state`.)

source/_patterns/00-atoms/forms/select.yaml
```diff
@@ -13,13 +13,28 @@ schema:
...
       type: string
       minLength: 1
-    state:
+    variant:
       type: string
       enum:
         - valid
         - error
+        - info
+    userInputInvalid:
```
thewilkybarkid: This is a helper for `state` == `error`? Seems ok given the limitations of Mustache. (Maybe call it `isError`? Or maybe `error` should be `invalid`, so `isInvalid`?)

source/_patterns/00-atoms/forms/text-field.mustache
```diff
@@ -9,18 +9,30 @@
...
     {{#disabled}} disabled="disabled"{{/disabled}}
     {{#placeholder}} placeholder="{{placeholder}}"{{/placeholder}}
+    {{#isInvalid}} aria-invalid="true"{{/isInvalid}}
+    {{#messageGroup}} aria-describedby="{{id}}"{{/messageGroup}}
   >
 
-  {{#message}}
-    <div class="form-item__message">{{{.}}}</div>
-  {{/message}}
+  {{#messageGroup}}
```
davidcmoulton: @thewilkybarkid swapping out `message` for `messageGroup`, separating the error message from the info message: none, either or both messages may be present. If you're happy with that approach from a data point of view I'll make the rest of the changes to support this here & in the view models.

source/_patterns/00-atoms/forms/text-field.yaml
```diff
@@ -17,9 +17,40 @@ schema:
...
+        - required:
+          - id
+          - infoText
+          not:
+            required:
+              - errorText
+        - required:
+          - id
+          - errorText
+          - infoText
```
thewilkybarkid: 🙃 

Might be clear to make `id` required and just have `minProperties: 2`. (Technically that means _any_ 2 properties, but meh.)

source/_patterns/00-atoms/forms/select.yaml
```diff
@@ -13,13 +13,30 @@ schema:
     name:
       type: string
       minLength: 1
-    state:
+    variant:
```
thewilkybarkid: Revert? Template still uses `state`.

source/_patterns/00-atoms/forms/select.yaml
```diff
@@ -13,13 +13,30 @@ schema:
...
       enum:
         - valid
-        - error
-    message:
-      type: string
+        - invalid
+    isInvalid:
+      type: boolean
+      enum:
+        - true
```
thewilkybarkid: Could remove this and just use `errorText`?

PR #756, opened 2017-09-15
PI-503-equations-in-table-footnotes

davidcmoulton:
Only set line-height on table caption descendants if they're not controlled by MathJax.


thewilkybarkid:
This does still override the font etc?

davidcmoulton:
All it touches is the line height.

PR #758, opened 2017-10-03
ELPP-2983 - profile page - login control

davidcmoulton:


davidcmoulton:
When the login control is finalised, the patterns involved in including it in the site header must be updated accordingly

.gitignore
```diff
@@ -20,3 +18,5 @@ test/build
 
 /vendor/
 /test/fixtures-local/
+/.idea/
```
thewilkybarkid: I'd keep stuff like this to your global `.gitignore`.

source/_patterns/01-molecules/navigation/profile-login-control.json
```diff
@@ -0,0 +1,7 @@
+{
+  "button": {
+    "text": "Login / Register",
```
thewilkybarkid: `Log in` (`Login` is a noun).

source/_patterns/01-molecules/navigation/profile-login-control.yaml
```diff
@@ -0,0 +1,30 @@
...
+      minLength: 1
+    logoutUri:
+      type: string
+      minLength: 1
+    displayName:
+      type: string
+      minLength: 1
+    button:
+      $ref: ../../00-atoms/components/button.yaml#/schema
+#   This dependencies block isn't working as expected: validation passes when isLoggedIn but not logoutUri is present
```
thewilkybarkid: `dependencies` is indented too much.

source/_patterns/01-molecules/navigation/profile-login-control.mustache
```diff
@@ -0,0 +1,15 @@
+<div class="profile-login-control" {{#isLoggedIn}}data-profile-home-link="{{profileHomeLink}}"
+     data-profile-manage-link="{{profileManageLink}}" data-logout-uri="{{logoutUri}}" data-display-name="{{displayName}}"{{/isLoggedIn}}
```
thewilkybarkid: Is there anyway we could decouple what appears in the menu from the template? (So we could add a new user-specific page in Journal without having to change the pattern.)

source/_patterns/04-pages/article--magazine.json
```diff
@@ -90,10 +90,21 @@
         {
           "text": "Submit my research",
           "path": "#",
-          "classes": "nav-secondary__item nav-secondary__item--button",
+          "classes": "nav-secondary__item",
```
thewilkybarkid: Do we need `classes` as a property anymore?

test/profilelogincontrol.spec.js
```diff
@@ -0,0 +1,394 @@
...
+        'shouldOnlyContainCharactersUseableWithinJavaScriptNames;plus-hyphens, commas, andSpaces',
+        'shouldOnlyContainCharactersUseableWithinJavaScriptNames:plus-hyphens, commas, andSpaces',
+        'shouldOnlyContainCharactersUseableWithinJavaScriptNames/plus-hyphens, commas, andSpaces',
+        'hyphens-should-not-be-at-the-end-of-the-string-',
+        'hyphens-should-not-be-at-the-end-of-a-word-, in-the-string',
+        '-hyphens-should-not-be-at-the-start-of-the-string',
+        'hyphens-should-not-be-at-the-start-of-a, -word-in-the-string'
+      ];
+
+      validDataAttributeRoots = [
```
davidcmoulton: Aim to get a regex fitting the final version of these into the json schema, for extra stringency.

thewilkybarkid:
Are we able to break the profile icon part into an image (with CSS rounded borders) and the arrow? The image _could_ be data too, so Journal would set it as the current one, and when we allow avatars there's no pattern change?

source/_patterns/04-pages/article--magazine.json
davidcmoulton: In some cases, yes (see https://github.com/elifesciences/pattern-library/blob/develop/source/_patterns/04-pages/article--magazine.json#L45). The more general case (primary/secondary) might be simplified through refactoring, but I think that's a separate task.

davidcmoulton:
We absolutely could, but I don't think that's MVP for first iteration.

davidcmoulton:
On investigation, it turns out that the image path inside the js file is not being rewritten correctly in https://github.com/elifesciences/journal/blob/develop/gulpfile.js#L139. To work around this, now including icon data as data attributes because, as @thewilkybarkid says, we will need it anyway later.

source/_patterns/01-molecules/navigation/login-control.mustache
```diff
@@ -0,0 +1,31 @@
...
+     data-link-field-roots="{{linkFieldRoots}}"
+     {{{linkFieldData}}}
+     data-default-uri="{{defaultUri}}"
+     data-display-name="{{displayName}}"
+
+     {{#icon}}
+       data-icon-path-root="{{pathRoot}}"
+       data-icon-srcset="{{srcset}}"
+       data-icon-alt-text="{{altText}}"
+     {{/icon}}
```
thewilkybarkid: In Carouseltem we embed the `<picture>` (https://github.com/elifesciences/pattern-library/blob/develop/source/_patterns/01-molecules/components/carousel-item.mustache#L3-L19)

davidcmoulton: I've never been totally happy with that approach, but it's the carousel, so let it go. In this case I think it's worth trying to get the data provided by journal and used by pattern-library (have always got the Carouseltem approach to fall back on if we need to).

assets/js/components/LoginControl.js
```diff
@@ -176,12 +176,19 @@ module.exports = class LoginControl {
...
+   * @param {HTMLElement} $elm pattern-root element
    * @param {Function} buildElement Function used to build an element (elife-utils.buildElement)
    * @return {HTMLElement} The resulting control element
    */
-  buildControl(extraLinksToBuild, buildElement) {
+  buildControl(extraLinksToBuild, $elm, buildElement) {
+    const iconData = {
+      pathRoot: $elm.dataset.iconPathRoot,
+      srcset: $elm.dataset.iconSrcset,
+      altText: $elm.dataset.iconAltText
```
davidcmoulton: `sizes`

source/_patterns/01-molecules/navigation/login-control.mustache
thewilkybarkid: I'm not that big a fan either, but it does allow us to use enable things like WebP (through `<source>`).

assets/js/components/LoginControl.js
```diff
@@ -0,0 +1,353 @@
...
+   * @returns {HTMLElement} the <picture> describing the icon
+   */
+  static buildIcon(iconData, buildElement) {
+    if (!(iconData.pathRoot && iconData.srcset && iconData.altText)) {
+      return null;
+    }
+
+    const $icon = buildElement.call(null, 'picture');
+    const $source = buildElement.call(null, 'source', [], '', $icon);
+    $source.setAttribute('srcset', `${iconData.pathRoot}.svg`);
```
thewilkybarkid: Journal should just provide a complete URI. User-loaded avatars won't (probably) be SVG.

source/_patterns/01-molecules/navigation/login-control.mustache
davidcmoulton: Good point. Okay, will go with the CarouselItem approach and see how it plays.

PR #770, opened 2017-10-30
Use pixel density srcset

davidcmoulton:
Same intention as, and supercedes #729


source/_patterns/00-atoms/components/picture.yaml
```diff
@@ -34,7 +34,7 @@ schema:
       properties:
         srcset:
           type: string
-          pattern: ^((https)?[^ ]+ [1-9][0-9]*w)(, (https?)?[^ ]+ [1-9][0-9]*w)*$
+          pattern: ^((https)?[^ ]+ [1-9][0-9]*x)(, (https?)?[^ ]+ [1-9][0-9]*x)*$
```
thewilkybarkid: Maybe reverse this and make the last one `1x`?

thewilkybarkid: The optional one should allow for decimals too

PR #774, opened 2017-11-03
Line height mixin

davidcmoulton:
This introduces the new mixin `font-size-and-vertical-height` that is to be used from now on to set all the line heights within the style blocks. It should also be used within mixins unless that doesn't make sense.

This means that every time a line height is set, the font size must now be specified too. This will help prevent bugs which can occur when the line height may be set independently of the font size, and the font size is computed by the browser is different from that which the developer expected.

Also, all uses of the bare number 16 for font sizing have been replaced by the sass variable `$font-size-base-in-px`.

### DIFF
--- a/assets/sass/_mixins.scss
+++ b/assets/sass/_mixins.scss
@@ -263,6 +263,12 @@
   }
 }
 
+// $vertical-height-in-px: desired height of the element as a function of font-size and line-height
+@mixin font-size-and-vertical-height($font-size, $vertical-height-in-px: 24) {
+  @include font-size($font-size);
+  line-height: #{($vertical-height-in-px / $font-size)};
+}
+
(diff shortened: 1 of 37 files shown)


assets/sass/_mixins.scss
```diff
@@ -398,22 +404,19 @@
 }
 
 @mixin popup_title_text {
-  @include font-size(14);
-  line-height: 1.4285714286; /* 20px on 24px font */
+  @include font-size-and-vertical-height(14, 20);
```
thewilkybarkid: Are we able to parameterise most/all of these values? They're a bit magic otherwise.

davidcmoulton: A lot have been, but more should be able to be, yes.

PR #775, opened 2017-11-06
Create new button variant login containing orcid icon, replacing confirm

davidcmoulton:


thewilkybarkid:
The PNG logo looks less clear to me? (SVG probably hasn't actually changed either?)

davidcmoulton:
This uses the 1x png, not the 2x: if the client can't view svg, it's also unlikely to have retina display.

Nick sent me a new SVG, so using that.

PR #780, opened 2017-11-16
Update package.json

jure:
This allows use of this repository as an NPM dependency.


davidcmoulton:
@jure how fixed is that name once it's in, if we change it later how much pain in there is accommodating that?

jure:
Not a problem, should be a one line change.

PR #787, opened 2017-11-24
Add new speech bubble variant to bubble pattern (for hypothesis)

davidcmoulton:


thewilkybarkid:
When we add this should JS only?

assets/sass/patterns/atoms/buttons.scss
```diff
@@ -85,5 +85,55 @@
...
+    @include colors($color-primary);
+    display: block;
+    font-family: $font-primary;
+    @include font-size-and-vertical-height(48, 36);
+    @include height(36);
+    @include padding(12 0 0);
+    position: relative;
+    @include width(42);
+
+    &:after {
```
thewilkybarkid: Is it possible to make this clickable too?

davidcmoulton: I've played with some hacky approaches and the answer is yes in theory, but want to find a nicer approach to it.

davidcmoulton:
@thewilkybarkid for this useage of it, js is required (getting to that), but it's conceived as a new button variant, so shouldn't be constrained to be js-only per se.

PR #792, opened 2017-12-11
Use getElementById() where possible

thewilkybarkid:
Hypothesis can share URIs like https://elifesciences.org/articles/22901#annotations:Rpp6JttuEeeXnRsvD_86eQ. The colon character needs to be escaped if using `querySelector`, otherwise there's a `Failed to execute 'querySelector' on 'Document': '#annotations:Rpp6JttuEeeXnRsvD_86eQ' is not a valid selector` exception.

This changes it to use `getElementById()` where possible. I haven't added escaping in the remaining cases as they (should) be under our control.

### DIFF
--- a/test/fragmenthandler.spec.js
+++ b/test/fragmenthandler.spec.js
@@ -44,11 +44,7 @@ describe('A FragmentHandler Component', function () {
               return [];
             }
           },
-          querySelector: function (selector) {
-            if (selector.indexOf('#') > -1) {
-              return true;
-            }
-          }
+          getElementById: function () {}
         };
 
         expect(fragmentHandler.getIdOfCollapsedSection('arbitraryElementId',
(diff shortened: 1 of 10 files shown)


test/fragmenthandler.spec.js
```diff
@@ -94,10 +90,31 @@ describe('A FragmentHandler Component', function () {
...
+              return [sectionMock]
+            }
+          },
+          getElementById: function (id) {
+            if(id === descendantId) {
+              return descendantMock;
+            }
+          }
+        };
+        let valueUnderTest = fragmentHandler.getIdOfCollapsedSection(descendantId, docMock);
```
thewilkybarkid: This JS testing hurts my head a bit. The test here was passing in a `HTMLElement` rather than a `HTMLDocument` as required anyhow.

PR #795, opened 2017-12-15
ELPP-3305 Don't hard-code investor logos

thewilkybarkid:


thewilkybarkid:
The `max-width` isn't working on IE11, and the widths go pretty weird on IE10.

source/_patterns/01-molecules/components/investor-logos.mustache
```diff
@@ -1,27 +1,24 @@
+<ol class="investor-logos">
+  {{#logos}}
+    <li class="investor-logos__item">
 
-<section class="investor-logos">
+      <div class="investor-logos__container">
+        <picture class="investor-logos__picture">
```
davidcmoulton: Might consider using `sizes` within here, as each footer image is a known quantity.

assets/sass/patterns/molecules/investor-logos.scss
```diff
@@ -1,63 +1,60 @@
...
+  list-style: none;
+  margin: 0;
+  padding-left: 0;
 }
 
-.investor-logos__img_container {
+.investor-logos__item {
   border-top: 1px solid $color-text-dividers;
-  @include height(120);
+  @include padding($padding);
```
davidcmoulton: `blg` (aka baseline grid) variables should only be used to set vertical space, not horizontal space: if the baseline grid changes, the horizontal spacing should be unaffected.

source/_patterns/01-molecules/components/investor-logos.mustache
davidcmoulton: The `yaml` states this is a `picture` object, but the markup is placed directly in here rather than included via a partial. Might be a bit confusing?

thewilkybarkid: We use this technique elsewhere when classes are required.

assets/sass/patterns/molecules/investor-logos.scss
```diff
@@ -1,63 +1,62 @@
...
 }
 
-.investor-logos__img_container {
+.investor-logos__item {
   border-top: 1px solid $color-text-dividers;
-  @include height(120);
+  @include blg-spacing("top", "small");
+  @include blg-spacing("bottom", "small");
+  @include padding(24, "left");
+  @include padding(24, "right");
```
thewilkybarkid: Won't do it here, but it would probably be a good idea to add `vertical` and `horizontal` as options.

source/_patterns/01-molecules/components/investor-logos.mustache
davidcmoulton: Ok, will keep this for consistency.

PR #796, opened 2017-12-19
ELPP-3284 annotation listing patterns

davidcmoulton:
- Add new pattern `annotation-teaser`
- Remove redundant teaser css
- Add and apply new mixin `truncate-with-ellipsis()`


source/_patterns/01-molecules/teasers/annotation-teaser.mustache
```diff
@@ -0,0 +1,34 @@
...
+
+  {{#highlight}}
+    <blockquote class="annotation-teaser__highlight">
+      <a href="{{url}}">{{{.}}}</a>
+    </blockquote>
+  {{/highlight}}
+
+  {{#content}}
+    <div class="annotation-teaser__content">
+      <a href="{{url}}">{{{.}}}</a>
```
thewilkybarkid: Are we meant to just strip any links in the annotation?

davidcmoulton: I understand from @nlisgo that no links within annotations will be carried over.

nlisgo: The annotation content could contain a link. Apologises for any confusion. There will certainly be instances of url's that are not links that the hypothesis client will render as links but we will not have to but there will also be instances where the markdown that is converted to content blocks will contain links. It is appropriate for the `annotations` api to present these as links, we should chat to @NickDuf about whether we should strip those links out for this presentation.

thewilkybarkid: Added https://github.com/elifesciences/journal/pull/820/commits/f4cd2dfc001dad31e02b015587c056669f034549 for now.

thewilkybarkid: We might want this link to be added by JavaScript too, since Hypothesis won't work for non-JS users. (Maybe JS is the right place to strip any links then?)

PR #798, opened 2018-01-02
ELPP-3329 Update contextual data

thewilkybarkid:
Not that happy with the implementation yet, but think it fulfils the brief.


davidcmoulton:
These additional commits add a small speech bubble variant, enabling the hypothesis opener pattern to be used for the annotation count instance within the contextual data pattern, in addition to its pre-existing usage in the main body of an article. The contextual data pattern is updated accordingly.

thewilkybarkid:
Implemented through to Journal now (https://github.com/elifesciences/journal/pull/835).

Couple of comments:

* Annotations link is now in two parts, which looks a bit funny (which is what I was trying to avoid, ie turned the `<dl>` into a `<ul>`).
* The Hypothesis Opener feels a bit over-worked. I think the real problem is the speech-bubble button: there's now a variant of a variant (`speech-bubble-small`). I'd suggest decoupling the opener from the button pattern (we don't need this speech bubble anywhere else?), or at least reuse the existing button sizing.

davidcmoulton:
@thewilkybarkid  Significant changes in these commits:
- `contextual-data` now has no cursor change when hovering between the word "Annotations" and the adjacent speech bubble.
- the speech bubble is now its own pattern, with behaviour for updating it. A `SpeechBubble` instance is used in the `HypothesisOpener` behaviour, avoiding the need for `HypothesisOpener` to reach in `speech-bubble` markup directly.
- the hypothesis opener no longer has any markup: it is almost behaviour-only apart from a little associated css. Note that `speech-bubble.yaml` calls this directly rather than via another yaml file.

source/_patterns/00-atoms/components/speech-bubble.yaml
```diff
@@ -0,0 +1,37 @@
...
+      type: boolean
+      enum:
+        - true
+    hasPlaceholder:
+      type: boolean
+      enum:
+        - true
+  required:
+    - text
+  anyOf:
```
thewilkybarkid: Any reason why you couldn't have both (even though we don't use it now)? Or doesn't it work?

source/_patterns/00-atoms/components/speech-bubble.json
```diff
@@ -0,0 +1,4 @@
+{
+  "text": "8",
+  "containsPlaceholder": true
```
thewilkybarkid: Needs to be removed?

source/_patterns/00-atoms/components/speech-bubble.yaml
```diff
@@ -0,0 +1,37 @@
+assets:
+  css:
+    - speech-bubble.css
+    - ../../../../assets/sass/patterns/molecules/hypothesis-opener.css
```
thewilkybarkid: Doesn't look right?

thewilkybarkid:
Re the cursor change, there's still a slight colour change depending on which bit you hover on.

I'm aware that we can't really hold this up much longer, but this still doesn't feel right. `SpeechBubble` doesn't look like it's _that_ reusable (eg the placeholder is tied to a quote mark), and I'm not sure should be separate at all?

source/_patterns/00-atoms/components/speech-bubble.yaml
davidcmoulton: No reason it shouldn't exist, but I haven't built the css for that as it's not required at the moment.

source/_patterns/00-atoms/components/speech-bubble.yaml
davidcmoulton: Yeah, I'm not very comfortable with this. It's like this because HypothesisOpener is now not a pattern, but still requires some positioning: this is calling in the css file for that directly rather than referring to a `yaml` file (because now there isn't one). Considered applying the style through JavaScript and so losing `hypothesis-opener.css` completely, the but that felt wrong. On reflection I think I might go for that option though, I think this is wronger.

davidcmoulton:
I agree. I think it's better than it was, but I'm still not particularly happy with this one.

I'm going to apply `hypothesis-opener.css` styles via JavaScript (ugh) so we can lose that orphan css file, and then I think it's time to call it for this iteration, as long as you're happy this approach doesn't store up trouble for the view model.

source/_patterns/00-atoms/components/speech-bubble.mustache
```diff
@@ -0,0 +1,4 @@
+<button class="speech-bubble{{#hasPlaceholder}} speech-bubble--has-placeholder{{/hasPlaceholder}}{{#isSmall}} speech-bubble--small{{/isSmall}}"
+        data-behaviour="SpeechBubble{{#behaviour}} {{behaviour}}{{/behaviour}}">
```
thewilkybarkid: `behaviour` isn't defined

PR #802, opened 2018-01-16
Update package json

davidcmoulton:
- adds a license field
- adds a repository field
- uses all lowercase characters in the name to conform to npm naming requirements

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,10 @@
 {
   "name": "eLife-pattern-library",
   "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/elifesciences/pattern-library"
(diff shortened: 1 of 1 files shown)


davidcmoulton:
The name change will affect clients who use the `package.json` `name` field for identification. @elifesciences/yld @yure let me know whether this will cause you problems.

tamlyn:
Fine with us. I think you meant to tag @jure.

jure:
Yes, this (`eLife-pattern-library` -> `elife-pattern-library`) is fine. Especially since the original name
 that I put in there is not a legal npm package name (must be all lower case https://docs.npmjs.com/files/package.json#name) 👍

PR #809, opened 2018-02-01
ELPP-3219 Make Button Collection a proper pattern

thewilkybarkid:
This is the long-way round getting a margin underneath buttons when they're part of the body content.

For future use (imagining buttons side by side), I decided to create a pattern that wrapped one or more buttons then add the margin to that, rather than try and add it to some buttons and not others (eg in the site header). Turns out that the Metrics pattern already had an informal pattern called Button Collection, so I've sorted that out.

I've also renamed the `button--outline` variant to `button--reversed`, then created a new variant called `button--outline` which is just an outline (not reversed as well!).

### DIFF
--- a/assets/sass/patterns/molecules/button-collection.scss
+++ b/assets/sass/patterns/molecules/button-collection.scss
@@ -1,27 +1,34 @@
 @import "../../definitions";
 
+$gutter: 10px;
+
 .button-collection {
   display: table;
+  @supports (display: flex) {
+    display: flex;
+    flex-wrap: wrap;
+    align-items: center;
+  }
   @include nospace();
 }
 
 .button-collection--centered {
   margin-left: auto;
   margin-right: auto;
+  padding-left: $gutter;
+  @supports (display: flex) {
+    justify-content: center;
+  }
 }
 
 .button-collection__item {
   float: left;
   list-style: none;
   @include blg-spacing("bottom", "small", "margin");
-  margin-right: 10px;
-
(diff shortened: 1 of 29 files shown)


assets/sass/patterns/molecules/button-collection.scss
```diff
@@ -2,42 +2,37 @@
...
-.button-collection__button--active {
-  background-color: $color-primary;
-  color: white;
 
-  &:hover {
-    background-color: $color-primary;
-    color: white;
+  .button-collection--compact & {
+    &:not(:first-child) {
+      .button {
```
thewilkybarkid: Getting pretty specific...

assets/sass/patterns/molecules/button-collection.scss
```diff
@@ -2,42 +2,37 @@
 
 .button-collection {
   display: table;
```
davidcmoulton: As this is being updated anyway, could use flexbox here?

assets/sass/patterns/molecules/carousel-item.scss
```diff
@@ -214,6 +214,9 @@
     bottom: 64px;
     left: 0;
     right: 0;
+    .button {
+      @include margin(0, "bottom");
```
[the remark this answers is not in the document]
davidcmoulton: Correction: `margin-bottom: 0;`

davidcmoulton:
👍

PR #812, opened 2018-02-06
ELPP-2643 Avoid variable resolution problems with link URLs

thewilkybarkid:
We often use the `name`/`url` properties combo for links and their text. Occasionally the link itself is optional, but `url` might exist at a higher level in the context and so it gets an incorrect value (eg context label in related item teasers should not link, but ends up with the `url` of the item itself).

This makes it so it has to be present but can be `false`. (Avoided a separate `hasUrl` property due to the schema complexity.)


source/_patterns/00-atoms/references/reference.yaml
```diff
@@ -33,10 +33,15 @@ schema:
                   type: string
                   minLength: 1
                 url:
-                  type: string
-                  minLength: 1
+                  oneOf:
```
giorgiosironi: not sure if it's possible to "extract" this `oneOf` type for reusability?

PR #813, opened 2018-02-07
Improve button collection margins

thewilkybarkid:
#809 implemented button collections with a bottom margin on each item to give space when the buttons are stacking, and for the overall bottom margin. However, because it's not on the list itself it doesn't collapse properly, so subsequent items have their entire top margin as well, such as:

screen shot 2018-02-07 at 13 39 44

This changes it to the only other method I can think of: put a top margin on each item, a _negative_ top margin on the list, then a normal bottom margin on the list. This should always result in the expected margins:

screen shot 2018-02-07 at 13 40 20


assets/sass/patterns/molecules/button-collection.scss
```diff
@@ -1,21 +1,23 @@
 @import "../../definitions";
 
-$gutter: 10px;
+$gutter: 10;
 
 .button-collection {
   display: table;
+  @include nospace();
+  @include blg-spacing("bottom", "small", "margin");
+  @include margin(-$blg-space-small-in-px, "top");
```
davidcmoulton: With all the hyphens in variable names, `@include margin($blg-space-small-in-px * -1, "top");` is easier to read here.

PR #821, opened 2018-02-21
Remove bottom padding from content headers when there's an image

thewilkybarkid:
This adds to the set height, which isn't meant (see https://github.com/elifesciences/journal/pull/831#issuecomment-367246813). The content is all positioned anyway, so it's also redundant.


thewilkybarkid:
Looking at it, it seems that bottom padding only applies to the read-more header anyway, normal ones have a `wrapper` class which sets it to 0...

PR #829, opened 2018-03-01
PI-640 Fix figure asset links when hidden

thewilkybarkid:


assets/js/components/AssetNavigation.js
```diff
@@ -61,6 +61,20 @@ module.exports = class AssetNavigation {
           this.addPreviousButton(i, navigation);
           this.addNextButton(i, navigation);
 
+          [].forEach.call(assetItem.querySelectorAll('a'), (link) => {
```
thewilkybarkid: Needs to happen outside `AssetNavigation`, could still apply if there's no supplements? (eg link to another figure's supplement)

assets/js/components/AssetNavigation.js
```diff
@@ -75,6 +75,7 @@ module.exports = class AssetNavigation {
     const previous = utils.buildElement('span', ['asset-viewer-inline__previous'], '', navigation);
     if (i > 0) {
       previous.addEventListener('click', () => {
+        this.window.history.replaceState({}, '', '#' + this.assetItems[i - 1].id);
```
thewilkybarkid: Avoids an edge case where the current hash is where the link is going (and so no `hashchange` event), eg you started on a supplement, went to the figure figure, clicked a link in the primary figure to go back to the supplement.

assets/js/components/AssetNavigation.js
```diff
@@ -75,6 +75,7 @@ module.exports = class AssetNavigation {
     const previous = utils.buildElement('span', ['asset-viewer-inline__previous'], '', navigation);
     if (i > 0) {
       previous.addEventListener('click', () => {
+        this.window.history.pushState(null, '', '#' + this.assetItems[i - 1].id);
```
thewilkybarkid: Avoids an edge case where the current hash is where the link is going (and so no hashchange event), eg you started on a supplement, went to the figure figure, clicked a link in the primary figure to go back to the supplement.

thewilkybarkid: This means the navigation always changes the fragment. For this actual problem, we could only care if the fragment is the ID of the asset that you're leaving, and remove it.

PR #830, opened 2018-03-07
ELPP-3498 Declare pattern JavaScript dependencies

davidcmoulton:
This PR declares the patterns' JavaScript dependencies within the `yaml` file in the same way that the CSS dependencies are declared. This includes a containing pattern containing references to its component patterns' JavaScript.


davidcmoulton:
Trying this out downstream fails the tests in `patterns-php` as the files aren't explicitly provided within the pattern library yet.

The `gulp` task could be amended accordingly, but this raises questions of how to handle transpilation. It's probably worth putting this on hold for now and continuing it as part of the bundling and transpilation review work we'll need to do when considering in more detail how we should implement asset delivery over HTTP/2.

PR #831, opened 2018-03-08
Commit statuses testing

giorgiosironi:
Just testing some commit statuses integrations, nothing to see here.


giorgiosironi:
Will re-use tomorrow, then close.

PR #835, opened 2018-03-15
Build container images to serve the static website

giorgiosironi:
This implementation build a couple of Docker images:
- `elifesciences/pattern-library` (Node-based)
- `elifesciences/pattern-library_ui` (nginx-based, but using PHP for building phase)

The first image will be used as a base to extend with tests, while the second one can be used as-is to serve the static website (assuming nothing is missing from there).

Both images can be pushed, there's the potential for some projects to reuse some of this work for a faster [bin/update](https://github.com/elifesciences/patterns-php/blob/master/bin/update#L32).


Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# end potentially base image
+
+RUN apt-get update && apt-get install -y \
+    bzip2 \
+    g++ \
+    make \
+    ruby \
+    ruby-dev 
+
+RUN gem install compass
```
thewilkybarkid: @davidcmoulton We really need to remove this since it's unsupported.

thewilkybarkid: Removes the Ruby constraint too.

giorgiosironi: :man_shrugging: I replicated what this is doing in production now

davidcmoulton: Agree. Ticket raised (ELPP-3514).

README.md
```diff
@@ -95,6 +95,26 @@ gulp test:selenium
...
+# Docker setup
+
+```
+docker build -t elifesciences/pattern-library .
+```
+
+(re)builds the Gulp-based Docker image, which builds all assets.
+
+```
+docker build -f Dockerfile.ui -t elifesciences/pattern-library_ui .
```
thewilkybarkid: Just thinking about naming, might be better to call this one `pattern-library` and the other `pattern-library_assets`.

giorgiosironi: Yes I think the primitive container with gulp is not much use in itself, wasn't sure that `pattern-library` always refers to a UI originally but it seems entrenched in the workflow (you modify code, you look at a browser).

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+COPY --chown=elife:elife source/ ${PROJECT_FOLDER}/source
+COPY --chown=elife:elife \
+    .babelrc \
+    .jscsrc \
+    .jshintrc \
+    .stylelintrc \
+    config.rb \
+    gulpfile.js \
+    ${PROJECT_FOLDER}/
+RUN gulp
```
thewilkybarkid: To use this in `patterns-php` we'll need the `--environment production` switch. Would that have to separately run as a `CMD`?

giorgiosironi: A `CMD` would execute it at runtime while in `patterns-php` we probably want to build another image from which we can extract files to commit. It seems it would only make sense to publish to Docker Hub an image where `--environment production` is used so maybe this option should be a build argument that defaults to true and can be turned off locally?

thewilkybarkid: It could be the default. Thinking about a container for local development, it'd need something like the `bin/dev` script that sets up multiple watches and runs a server, so probably needs to be a standalone.

PR #837, opened 2018-03-20
ELPP-3404 Hypothesis loading state indicators

davidcmoulton:


thewilkybarkid:
- There's a small jump (it shrinks) when loaded.
- The in-page opener has the empty text and loading indicator visible at the same time.
- The in-page opener's loading indicator isn't styled correctly.

thewilkybarkid:
Though I'm not sure the loading state is meant to apply to the in-page opener.

assets/js/components/HypothesisLoader.js
```diff
@@ -17,22 +17,26 @@ module.exports = class HypothesisLoader {
     try {
       this.window.hypothesisConfig = HypothesisLoader.assembleHypothesisConfig(this.window);
     } catch (e) {
-      console.error('Could\'t assemble valid Hypothesis config from external data, aborting Hypothesis client loading.');
-      console.error('The error was: ', e);
+      this.window.console.error(e);
+      if (typeof this.window.newrelic === 'object') {
```
davidcmoulton: Considered replacing this with a new utils function called something like `logToRUM` (abstracting out the New Relic specific code into it), but the feature in this PR should ship first. The next time Real User Metric logging is added would be the time to do that, and refactor this to use it.

thewilkybarkid:
This appears not to work, Hypothesis loads correctly but the loader continues and times out, disabling it:

davidcmoulton:
Investigation shows the problem to be that when running `journal` locally, Hypothesis is now not updating the `[data-hypothesis-annotation-count]` HTML element. The mutation of this element is used to indicate that Hypothesis has loaded and is meaningfully interactive, and so it is also used to cancel the timeout. This is a recent regression: it was working fine a week ago.

Running `journal` locally with the currently live pattern codebase shows the same situation: the contents of `[data-hypothesis-annotation-count]` is not updated, and so the current comment count is never displayed. However, it's working fine on the live site, suggesting this is probably an issue with Hypothesis on the eLife test authority rather than explicitly with the functionality of this feature as written.

That said, this manifestation highlights the brittleness of the timeout approach.  Should for whatever reason the comment count fail to be updated, but the Hypothesis tool is anyway loaded and interactive, the current implementation would prevent its use. I'll investigate alternative approaches with the product team.

davidcmoulton:
@thewilkybarkid You need to access `journal` on a specific url to use the eLife test authority locally. I'll send you the details.

thewilkybarkid:
Confirmed works when configured correctly (and isn't worth considering here).

When triggering a load failure the cursor isn't entirely correct, still appears as clickable in-between the text/icon and to the right of the icon:

screen shot 2018-05-02 at 08 34 07

source/_patterns/00-atoms/components/speech-bubble.yaml
```diff
@@ -8,7 +8,6 @@ schema:
   properties:
     text:
       type: string
-      minLength: 1
```
thewilkybarkid: Needed?

PR #838, opened 2018-03-26
ELPP-3436: Avoid orphan further reading heading

davidcmoulton:


source/_patterns/02-organisms/components/listing-read-more.yaml
```diff
@@ -10,6 +10,10 @@ schema:
   $schema: http://json-schema.org/draft-04/schema#
   type: object
   properties:
+    allRelated:
+      type: boolean
+      emum:
+        - true
```
thewilkybarkid: Redundant (not being there is the same as `false`).

source/_patterns/02-organisms/components/listing-read-more~all-related.json
```diff
@@ -0,0 +1,67 @@
+{
+  "heading": {
+    "heading": "Further reading"
+  },
+  "items": [
```
thewilkybarkid: Meant to have `allRelated: true`?

PR #840, opened 2018-04-03
ELPP-3500 Create static site on PR from container image

giorgiosironi:
The end result is the same: https://s3.amazonaws.com/ci-pattern-library/840/index.html
It is created from the container image rather than from the VM.

The smoke tests also target a `ui` container rather than the node, so that we can remove the `pattern-library--ci` VM from the pipeline altogether, cutting 2 minutes from every build.


giorgiosironi:
Need to provision AWS credentials on the `containers` nodes.

Jenkinsfile
```diff
@@ -15,36 +15,37 @@ elifePipeline {
...
                 sh "docker-compose down -v"
             }
-        },
-        'containers--medium'
-    )
-
-    stage 'Smoke tests', {
-        lock('pattern-library--ci') {
-            builderDeployRevision 'pattern-library--ci', commit
-            builderSmokeTests 'pattern-library--ci', '/srv/pattern-library'
```
thewilkybarkid: This is the cause of losing the separate `smoke_tests` PR status?

giorgiosironi: Yes, because the new smoke tests are implemented manually to show the docker commands used. I'll wrap them into the Github commit status "context manager".

Jenkinsfile
```diff
@@ -13,38 +13,43 @@ elifePipeline {
...
 
-    stage 'Smoke tests', {
-        lock('pattern-library--ci') {
-            builderDeployRevision 'pattern-library--ci', commit
-            builderSmokeTests 'pattern-library--ci', '/srv/pattern-library'
-            // it is not yet possible to retrieve a JUnit XML log to archive as a test artifact:
-            // - the `xunit` formatter mangles the XML outputting also debug statements between tags
-            // - the `xunit-file` formatter, which is an external plugin, doesn't seem to work with gulp-mocha-phantomjs
+            stage 'Smoke tests', {
+                withCommitStatus({
```
thewilkybarkid: Doesn't seem to cause the links to go to the right step, but not that important.

giorgiosironi: The manually generated commit statuses always to go [`env.RUN_DISPLAY_URL`](https://github.com/elifesciences/elife-jenkins-workflow-libs/blob/master/vars/withCommitStatus.groovy), but in case of failures Jenkins may open the failed step.

PR #841, opened 2018-04-04
ELPP-3573 Resolve content header list wrapping

thewilkybarkid:
Main goal is to stop the corresponding author icon from wrapping:

screen shot 2018-04-04 at 08 33 54

But I've also spotted that the expander can wrap:

screen shot 2018-04-04 at 08 34 21

And the last-non-excess institution misses its separator:

screen shot 2018-04-04 at 08 34 35

Making the icon not wrap isn't hard, but the CSS-introduced comma is problematic. So, I'm thinking that it should be in the HTML (as ARIA hidden) and shown/displayed as appropriate.

(I've applied it to institutions as well even though they don't have the icon and can probably get away with it, for consistency.)


giorgiosironi:
I'll merge the latest `develop` into this to avoid depending `pattern-library--ci` for further builds.

source/_patterns/02-organisms/content-headers/content-header.mustache
```diff
@@ -121,13 +121,15 @@
       <ol class="content-header__author_list">
         {{#authors.list}}
           <li class="content-header__author_list_item">
-            <span class="content-header__author">{{#url}}<a href="{{url}}" data-behaviour="Popup" class="content-header__author_link">{{/url}}{{name}}{{#url}}</a>{{/url}}{{#isCorresponding}}
-              <picture>
+            <span class="content-header__author">{{#url}}<a href="{{url}}" data-behaviour="Popup" class="content-header__author_link">{{/url}}{{name}}{{#url}}</a>{{/url}}{{#isCorresponding}}&nbsp;{{!
```
thewilkybarkid: Can still recreate the original problem, looks like `&nbsp;` doesn't fix non-text?

thewilkybarkid: https://stackoverflow.com/questions/25041857/why-do-inline-blocks-break-after-non-breaking-space

thewilkybarkid:
@giorgiosironi The S3 deploy doesn't seem to have the changes. I though #847 might solve the problem but apparently not.

thewilkybarkid:
Seems that the problem is https://github.com/elifesciences/pattern-library/blob/98748e54c93ae893bb21d46ca8c3ae15911121d5/Dockerfile#L25-L27 where it's not recognising changes.

docker-compose.yml
```diff
@@ -13,6 +13,8 @@ services:
         build:
             context: .
             dockerfile: Dockerfile
+            args:
+                commit: ${IMAGE_TAG}
```
thewilkybarkid: @giorgiosironi This seems to be the fix (not setting means it always gets the `latest` assets).

thewilkybarkid: Extracted into #849.

PR #844, opened 2018-04-09
ELPP-3500 Push assets image

giorgiosironi:
Necessary for `patterns-php` integration.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,6 +12,10 @@ elifePipeline {
                 checkout scm
                 sh "IMAGE_TAG=${commit} docker-compose build"
                 image = DockerImage.elifesciences(this, "pattern-library", commit)
+                elifePullRequestOnly { prNumber ->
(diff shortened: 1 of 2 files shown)


giorgiosironi:
See https://hub.docker.com/r/elifesciences/pattern-library_assets/tags/

Jenkinsfile
```diff
@@ -12,6 +12,10 @@ elifePipeline {
                 checkout scm
                 sh "IMAGE_TAG=${commit} docker-compose build"
                 image = DockerImage.elifesciences(this, "pattern-library", commit)
+                elifePullRequestOnly { prNumber ->
+                    def assetsImage = DockerImage.elifesciences(this, "pattern-library_assets", commit)
+                    assetsImage.tag("pr-${prNumber}").push()
```
thewilkybarkid: Might be ok to do this before the tests so we can use the PR even if it's not yet green. Maybe add a comment to say this is intended.

PR #845, opened 2018-04-09
ELPP-3500 Optimize docker-compose build

giorgiosironi:
`ENVIRONMENT=development docker-compose build assets` down to:
- <1 second for no change
- 12 seconds for a CSS change
- ~~24~~ 20 seconds for a JavaScript change

The main bottleneck tasks are:
- `buildStyleFiles` > 6 seconds
- `js:hint` > 1.5 seconds
- `js:cs` > 1.5 seconds
- `js` > ~~6~~ 3 seconds

### DIFF
--- a/.env
+++ b/.env
@@ -1,2 +1,3 @@
 COMPOSE_PROJECT_NAME=pattern-library
 IMAGE_TAG=develop
+ENVIRONMENT=production
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,16 +15,16 @@ COPY --chown=elife:elife \
 # customized command for composer
 RUN composer --no-interaction install
 
-COPY --chown=elife:elife bin/ ${PROJECT_FOLDER}/bin
+COPY --chown=elife:elife \
+    core/styleguide ${PROJECT_FOLDER}/public/styleguide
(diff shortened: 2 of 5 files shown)


giorgiosironi:
Not sure if it's used for the local development cycle, but if we could turn off `sourcemap` https://github.com/elifesciences/pattern-library/blob/develop/gulpfile.js#L73 that shaves another 1-2 seconds off each change rebuild.

thewilkybarkid:
Subject to @davidcmoulton's approval, but I suggest we stop running `js:hint`, `js:cs`, `browserify-tests` and `sass:lint` here, as they're for testing/CI.

`generateIndividualStyles` could be dropped for now too, since we're not using/aren't going to use it in  that form (we'll end up with something similar, but more aggregated).

giorgiosironi:
> Possibly crazy thought, but could we split CSS/JS generation into separate containers, then assets is just the combination of their outputs?

Yes, although they would need a common base image to perform only one `npm install`, so it may get a bit more complex

giorgiosironi:
The current state of this does not change what is executed, can it go ahead while we think of more optimizations?

PR #846, opened 2018-04-10
Prevent FOIT

thewilkybarkid:
Rather than the text not appearing, this renders it in the fallback then re-renders it when the font is loaded.

We might want to consider `optional`, which won't re-render if the font is slow to load (but since it's _very_ cacheable it might be available on the next page load).


davidcmoulton:
👍

PR #847, opened 2018-04-10
Improve PR S3 deploy

thewilkybarkid:


Jenkinsfile
```diff
@@ -42,10 +42,10 @@ elifePipeline {
                     elifeGithubCommitStatus commit, 'pending', 'continuous-integration/jenkins/pr-demo', 'Static website is being built', url
 
                     def container = sh(script: "docker run -d elifesciences/pattern-library:${commit}", returnStdout: true).trim()
-                    sh "docker stop ${container}"
                     sh "docker cp ${container}:/usr/share/nginx/html public/"
+                    sh "docker stop ${container}"
                     sh "docker rm ${container}"
-                    sh "aws s3 cp public/ s3://ci-pattern-library/${prNumber}/ --recursive"
+                    sh "aws s3 sync public/ s3://ci-pattern-library/${prNumber}/ --delete"
```
thewilkybarkid: Currently always copies files as the timestamps changes. `--size-only` could be used, but might miss some small changes (ie when the size remains the same).

Jenkinsfile
```diff
@@ -42,10 +42,10 @@ elifePipeline {
                     elifeGithubCommitStatus commit, 'pending', 'continuous-integration/jenkins/pr-demo', 'Static website is being built', url
 
                     def container = sh(script: "docker run -d elifesciences/pattern-library:${commit}", returnStdout: true).trim()
+                    sh "docker cp ${container}:/usr/share/nginx/html/. public/"
                     sh "docker stop ${container}"
-                    sh "docker cp ${container}:/usr/share/nginx/html public/"
                     sh "docker rm ${container}"
-                    sh "aws s3 cp public/ s3://ci-pattern-library/${prNumber}/ --recursive"
+                    sh "aws s3 sync public/ s3://ci-pattern-library/${prNumber}/ --delete"
```
thewilkybarkid: Currently always copies files as the timestamps changes. `--size-only` could be used, but might miss some small changes (ie when the size remains the same).

giorgiosironi: like changing a color :thinking:

PR #849, opened 2018-04-10
Set commit argument

thewilkybarkid:
Fixed PR deploy in #841 and tests in #837 as they were both using `latest`.

Name of the argument probably needs changing too?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@ services:
             context: .
             dockerfile: Dockerfile
             args:
-                commit: ${IMAGE_TAG}
+                image_tag: ${IMAGE_TAG}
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -22,6 +24,8 @@ services:
         build:
             context: .
             dockerfile: Dockerfile.ci
+            args:
+                commit: ${IMAGE_TAG}
```
giorgiosironi: the alternative to the `latest` default is to have a `develop` default but it doesn't help much, only makes it more annoying to use the Dockerfile in isolation (and building should always go through `docker-compose build` to correctly refer to all upstream images)

PR #850, opened 2018-04-11
ELPP-3500 Push assets image tagged as the commit

giorgiosironi:
Was already pushed before, but only with the `latest` tag and we may want to pin this tag in `patterns-php`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -66,8 +66,6 @@ elifePipeline {
 
                 stage 'Approval', {
                     elifeGitMoveToBranch commit, 'approved'
-                    assetsImage.tag('approved').push()
-                    image.tag('approved').push()
                 }
             }
         },
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,9 +12,9 @@ elifePipeline {
 
             stage 'Build images', {
                 sh "IMAGE_TAG=${commit} docker-compose build"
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -11,12 +12,12 @@ elifePipeline {
 
             stage 'Build images', {
                 sh "IMAGE_TAG=${commit} docker-compose build"
-                image = DockerImage.elifesciences(this, "pattern-library", commit)
                 elifePullRequestOnly { prNumber ->
                     // push immediately to allow downstream exploration even with automated tests failing
-                    def assetsImage = DockerImage.elifesciences(this, "pattern-library_assets", commit)
+                    assetsImage = DockerImage.elifesciences(this, "pattern-library_assets", commit)
```
thewilkybarkid: Needs to be outside `elifePullRequestOnly`?

Jenkinsfile
```diff
@@ -58,12 +59,14 @@ elifePipeline {
...
             elifeMainlineOnly {
-                stage 'Push image', {
+                stage 'Push images', {
+                    assetsImage.push()
                     image.push()
                 }
 
                 stage 'Approval', {
                     elifeGitMoveToBranch commit, 'approved'
+                    assetsImage.tag('approved').push()
```
thewilkybarkid: Bit redundant.

giorgiosironi: Would drop the `approved` tag for both images?

PR #851, opened 2018-04-12
Only store the assets in the assets container

thewilkybarkid:
This should reduce the size, as we don't need to distribute the tools required to build the assets.

### DIFF
--- a/Dockerfile.assets
+++ b/Dockerfile.assets
@@ -1,40 +1,5 @@
-# Debian oldoldstable, support ends in May 2018
-# we need this for Ruby 1.9.x at the moment
-FROM node:6.13.1-wheezy AS assets-builder
-
-# TODO: base image
-USER root
-RUN useradd -ms /bin/bash -G www-data elife && \
-    chown elife:elife /srv && \
-    mkdir /srv/bin && \
-    chown elife:elife /srv/bin && \
-    mkdir -p /var/www && \
-    chown www-data:www-data /var/www
-ENV PATH=/srv/bin:${PATH}
-# end base image
-
-# potentially base image
-RUN npm install -g gulp-cli
-# end potentially base image
-
-RUN apt-get update && apt-get install -y \
-    bzip2 \
-    g++ \
-    make \
-    ruby \
-    ruby-dev 
-
-RUN gem install compass
-
-USER elife
-ENV PROJECT_FOLDER=/srv/pattern-library
-RUN mkdir ${PROJECT_FOLDER}
-WORKDIR ${PROJECT_FOLDER}
-COPY --chown=elife:elife \
-    package.json \
-    npm-shrinkwrap.json \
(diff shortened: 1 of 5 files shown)


giorgiosironi:
> CI which currently uses it for Node

If the `Dockerfile.assets` is split into `Dockerfile.assets-builder` and `Dockerfile.assets`, `Dockerfile.ci` can then extend the former?

thewilkybarkid:
The `assets` container is ok, and is down from 355MB to 2MB.

Dockerfile.assets
```diff
@@ -58,3 +23,12 @@ COPY --chown=elife:elife source/ ${PROJECT_FOLDER}/source
 
 COPY --chown=elife:elife assets/sass ${PROJECT_FOLDER}/assets/sass
 RUN gulp --environment ${environment} generateStyles
+
+FROM busybox:1.28.3
```
giorgiosironi: If this is going to be purely files without executables, you could try [`FROM scratch`](https://hub.docker.com/_/scratch/)

giorgiosironi: Although `docker-compose.yml` (being oriented to services) needs a command to execute so easier to have `sh` around

thewilkybarkid: Yeah, tried that first. Wasn't sure what to use as `command`.

docker-compose.yml
```diff
@@ -30,8 +33,15 @@ services:
         command: /bin/bash
         depends_on:
             - assets
+            - node
             - ui
             - selenium
+    node:
```
giorgiosironi: `node` should be at the top probably

Dockerfile.assets
```diff
@@ -58,3 +23,12 @@ COPY --chown=elife:elife source/ ${PROJECT_FOLDER}/source
 
 COPY --chown=elife:elife assets/sass ${PROJECT_FOLDER}/assets/sass
 RUN gulp --environment ${environment} generateStyles
+
```
giorgiosironi: shouldn't everything before this go into the `node` image? Then the `ci` image can use it (e.g. the `gulpfile.js` absence which the makes the build fail)

Dockerfile.assets
```diff
@@ -1,60 +1,11 @@
...
-
-COPY --chown=elife:elife assets/js/ ${PROJECT_FOLDER}/assets/js
-RUN gulp --environment ${environment} js
-
-COPY --chown=elife:elife source/ ${PROJECT_FOLDER}/source
 
-COPY --chown=elife:elife assets/sass ${PROJECT_FOLDER}/assets/sass
-RUN gulp --environment ${environment} generateStyles
+COPY --from=assets-builder \
+    --chown=root:root \
```
giorgiosironi: `--chown=root:root` should be the default of `COPY`

PR #852, opened 2018-04-16
ELPP-3500 Optimize browserify-tests

giorgiosironi:
This command should not be needed in the main `js` task, as it can be run when tests are actually needed (`gulp test`).

Moreover, with a separate `test:unit` target to be run only in the `ci` container, we can run only test and skip the `js` task that has already been executed during the building of the `assets` image.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -177,7 +177,7 @@ gulp.task('fonts', () => {
  * Creates a sourcemap.
  ******************************************************************************/
 
-gulp.task('js', ['js:hint', 'js:cs', 'browserify-tests'], () => {
+gulp.task('js', ['js:hint', 'js:cs'], () => {
 
     return browserify('./assets/js/main.js', { debug: true })
           .transform(babel)
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -233,7 +233,7 @@ gulp.task('browserify-tests', (done) => {
 });
 
 
-gulp.task('test', ['browserify-tests', 'js'], () => {
+gulp.task('local:test:unit', ['browserify-tests', 'js'], () => {
   return gulp.src('./test/*.html')
     .pipe(mochaPhantomjs({
       reporter: 'spec',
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -177,7 +177,7 @@ gulp.task('fonts', () => {
  * Creates a sourcemap.
  ******************************************************************************/
 
-gulp.task('js', ['js:hint', 'js:cs', 'browserify-tests'], () => {
+gulp.task('js', ['js:hint', 'js:cs'], () => {
```
giorgiosironi: @davidcmoulton correct me if I'm wrong, but it seems `js:hint` and `js:cs` do not take into account the contents of `test/`. If that's true, we don't need to run them separately in the `ci` image (which is the only one actually containing the tests), which would be great.

gulpfile.js
```diff
@@ -245,6 +245,18 @@ gulp.task('test', ['browserify-tests', 'js'], () => {
     .pipe(reload());
 });
 
+gulp.task('test:unit', ['browserify-tests'], () => {
```
thewilkybarkid: Duplicate of `test`? Maybe `test` should run `test:unit` and `test:selenium`?

giorgiosironi: I think `test` is still being used to run locally, in that case it would still need to run `js` to ensure code is up-to-date. Could renamed it to `local:test` or `local:test:unit`.

gulpfile.js
davidcmoulton: @giorgiosironi yes, `js:hint` and `js:cs` do not run over the `test/` directory (although whether they should is another question).

gulpfile.js
davidcmoulton: `test` is run locally during development. Happy for it to be renamed `local:test:unit` for clarity.

gulpfile.js
giorgiosironi: I'll keep the current behavior, checks could be run there if needed, but they would be different tasks so that they can be separated

PR #853, opened 2018-04-16
[donotmerge] ELPP-3500 Proof of concept: gulp watch with Docker

giorgiosironi:
There are two main ways to go about implementing watching tasks with our container images.

[TBC]

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,10 @@ services:
             dockerfile: Dockerfile.assets-builder
         image: elifesciences/pattern-library_assets-builder:${IMAGE_TAG}
         command: /bin/bash
+        user: "1001"
+        volumes:
+            - ./:/srv/pattern-library
+            - source_assets:/srv/pattern-library/source/assets
     assets:
         build:
             context: .
@@ -52,3 +56,6 @@ services:
         volumes:
             # https://github.com/SeleniumHQ/docker-selenium#running-the-images
             - /dev/shm:/dev/shm
+
+volumes:
+    source_assets:
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -177,7 +177,7 @@ gulp.task('fonts', () => {
  * Creates a sourcemap.
  ******************************************************************************/
 
-gulp.task('js', ['js:hint', 'js:cs', 'browserify-tests'], () => {
+gulp.task('js', ['js:hint', 'js:cs'], () => {
 
     return browserify('./assets/js/main.js', { debug: true })
           .transform(babel)


docker-compose.yml
```diff
@@ -7,6 +7,10 @@ services:
             dockerfile: Dockerfile.assets-builder
         image: elifesciences/pattern-library_assets-builder:${IMAGE_TAG}
         command: /bin/bash
+        user: "1001"
```
giorgiosironi: may not be necessary with the `source_assets` volume in place

gulpfile.js
```diff
@@ -177,7 +177,7 @@ gulp.task('fonts', () => {
  * Creates a sourcemap.
  ******************************************************************************/
 
-gulp.task('js', ['js:hint', 'js:cs', 'browserify-tests'], () => {
+gulp.task('js', ['js:hint', 'js:cs'], () => {
```
giorgiosironi: see https://github.com/elifesciences/pattern-library/pull/852 may also not be strictly necessary here

docker-compose.yml
```diff
@@ -7,6 +7,10 @@ services:
...
+        volumes:
+            - ./:/srv/pattern-library
+            - source_assets:/srv/pattern-library/source/assets
```
giorgiosironi: uh, have to "unmount" node_modules here, probably

giorgiosironi: @thewilkybarkid I don't remember where you used the syntax to allow a folder from the original image, inside a mounted volume?

giorgiosironi: (necessary to run this on the Mac test mule, as `npm install` on the host was failing)

thewilkybarkid: https://github.com/libero/walking-skeleton/blob/fa977652b53c55c34bb726203cc1ff90feaf5951/docker-compose.yml#L9-L11

giorgiosironi: any doc reference for that?

giorgiosironi:
`gulp watch` seems to work across time and space also using a Mac. I'll clean this up.

giorgiosironi:
Separated `dev` usage from `ci` usage, but there's the Ctrl+C recurring problem to solve.

docker-compose.yml
thewilkybarkid: Hmm, not really. Syntax is mentioned on https://docs.docker.com/compose/compose-file/#short-syntax-3. Can't remember where I found it.

giorgiosironi:
Also I'm not convinced about the behavior of the single string `volumes` key, have to experiment.

Jenkinsfile
```diff
@@ -34,10 +34,10 @@ elifePipeline {
 
             stage 'Smoke tests', {
                 withCommitStatus({
-                    sh "IMAGE_TAG=${commit} docker-compose run ci ./smoke_tests.sh ui http"
+                    sh "IMAGE_TAG=${commit} docker-compose -f docker-compose.yml run ci ./smoke_tests.sh ui http"
 
                     // preserve environment to allow investigation if build fails, clean up otherwise
-                    sh "IMAGE_TAG=${commit} docker-compose down -v"
+                    sh "IMAGE_TAG=${commit} docker-compose -f docker-compose.yml down -v"
```
giorgiosironi: to avoid using `docker-compose.override.yml` in `ci`

bin/gulp-watch
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+trap "echo ' ...Ctrl+C'; docker kill pattern-library-gulp-watch > /dev/null; docker rm pattern-library-gulp-watch > /dev/null" SIGINT
```
giorgiosironi: From what I can understand signals cannot be easily propagated from from `docker-compose run ...` to the internal `gulp` process, which runs with PID 1 and is under 3-4 layers of indirection. The best bet is to trap the signals and do an explicit kill&cleanup.

docker-compose.override.yml
```diff
@@ -0,0 +1,11 @@
+version: '3'
+
+services:
+    assets-builder:
+        volumes:
+            - ./:/srv/pattern-library
+            - /srv/pattern-library/node_modules
```
giorgiosironi: This doesn't think what we/I thought it does: it creates a volume for `node_modules`, and like most (all?) Docker volumes it copies the contents of that folder from the image into the volume. This being a builder container, it doesn't matter very much as downstream Dockerfiles will copy files from the `assets-builder` image, not from any volume. We also create a container in `bin/gulp-watch`, but it's a throwaway that gets re-created every time is needed (I should add `-v` to `docker rm`). But if it's a service it may cause problems as the volume is not updated when a new image is run...

giorgiosironi:
Updated the PR description and README, ready for review now.

docker-compose.override.yml
thewilkybarkid: Ah, https://stackoverflow.com/a/38601156/77199 mentions persistence.

docker-compose.override.yml
```diff
@@ -0,0 +1,11 @@
...
+            - source_assets:/srv/pattern-library/source/assets
```
thewilkybarkid: We're not expecting these to go anywhere yet, just demonstrating that `gulp watch` works?

thewilkybarkid: We're not expecting these to go anywhere yet, just demonstrating that gulp watch works?

PR #855, opened 2018-04-20
ELPP-3500 Watch with ui-builder as well as with gulp

giorgiosironi:
Propagates changes e.g. to `source/_patterns/` to the ui visible at http://localhost:8889

PR #856, opened 2018-04-24
ELPP-3588 Separate Composer and improve image

thewilkybarkid:
From https://github.com/elifesciences/pattern-library/pull/855#discussion_r183353630 (will conflict with #855, but easily resolvable).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,21 @@
 ARG image_tag=latest
-# looking for a better name for the node image that replaces `assets`
 FROM elifesciences/pattern-library_assets:${image_tag} AS assets
-FROM elifesciences/php_cli:22434ef5bda09326d4c9347de7d8c2f1610a0b83 AS ui-builder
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,34 +1,21 @@
 ARG image_tag=latest
-# looking for a better name for the node image that replaces `assets`
 FROM elifesciences/pattern-library_assets:${image_tag} AS assets
-FROM elifesciences/php_cli:22434ef5bda09326d4c9347de7d8c2f1610a0b83 AS ui-builder
+FROM elifesciences/pattern-library_composer:${image_tag} AS composer
+FROM php:7.0.29-cli-alpine AS ui-builder
```
thewilkybarkid: Deliberately not an eLife image since it doesn't need to be.

giorgiosironi: Probably not a problem for what it does (running a script in an isolated environment), it won't have the new configuration https://github.com/elifesciences/elife-base-images/blob/master/config/php-7.0-elife.ini

PR #861, opened 2018-04-27
ELPP-3500 Workflow support for bin/watch

giorgiosironi:
Various improvements from a morning session with @davidcmoulton.

Contains everything but `tests:watch`, which will be integrated separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,6 @@ This script does not allow to run tests on the result of the watch (yet).
 Exit from this script with `Ctrl+C`.
 
 TODO: add docker exec example to run arbitrary gulp commands
-TODO: mount source/assets/ as a folder from the host rather than as a volume
 TODO: mount test/build/ as a volume
 TODO: remove docker-compose build from bin/watch for better restart performance
 TODO: tests:watch should be bin/tests-watch, having / source folder as a volume
--- a/README.md
+++ b/README.md
@@ -152,11 +152,15 @@ You can pass options to the underlying gulp:
 bin/watch --sass-lint=false
 ```
 
-This script does not allow to run tests on the result of the watch (yet).
+To run additional gulp command in the same container where `gulp:watch` is running:
+
+```
+$ docker exec -it pattern-library-gulp-watch /bin/bash
+elife@...$ gulp test:unit
+```
 
 Exit from this script with `Ctrl+C`.
 
-TODO: add docker exec example to run arbitrary gulp commands
 TODO: tests:watch should be bin/tests-watch, having / source folder as a volume
 TODO: for tests:watch, forward 3000, 3001, 8080 to gulp container (or perhaps a different container)
 
--- a/README.md
+++ b/README.md
@@ -156,7 +156,6 @@ This script does not allow to run tests on the result of the watch (yet).
 Exit from this script with `Ctrl+C`.
 
 TODO: add docker exec example to run arbitrary gulp commands
-TODO: mount test/build/ as a volume
 TODO: remove docker-compose build from bin/watch for better restart performance
 TODO: tests:watch should be bin/tests-watch, having / source folder as a volume
 TODO: for tests:watch, forward 3000, 3001, 8080 to gulp container (or perhaps a different container)
--- a/README.md
+++ b/README.md
@@ -140,6 +140,7 @@ ENVIRONMENT=development docker-compose build assets
 To watch for changes, run:
 
 ```
+docker-compose build  # only necessary after switching branch or installing new dependencies
 bin/watch
 ```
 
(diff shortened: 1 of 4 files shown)


bin/extract-assets
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+docker cp pattern-library-gulp-watch:/srv/pattern-library/source/assets .container_source_assets
```
davidcmoulton: Could there be an option to run this as a watch task to keep `.container_source_assets` up to date with whatever's generated within the container?

giorgiosironi: The key thing here is separating writes performed by the container from the source owned by the host, so I'll try to change this into a mounted folder.

giorgiosironi:
`source/assets/` can't be a mounted host folder because it would start out as empty rather than with the default contents of the built image (which happens when it is a volume). Therefore, the last change keeps the mounted `.container_source_assets` up-to-date with it.

docker-compose.watch.yml
```diff
@@ -6,6 +6,7 @@ services:
             - ./:/srv/pattern-library
             - /srv/pattern-library/node_modules
             - source_assets:/srv/pattern-library/source/assets
+            - ./.container_source_assets:/srv/pattern-library/.container_source_assets
```
davidcmoulton: Changes to  `assets/js/main.js` on the host update within `.container_source_assets` as expected.

CSS does not update as expected. For example when editing `assets/sass/patters/molecules/audio-player.scss` on the host, the change does not propagate to `.container_source_assets/css/audio-player.css` nor to `.container_source_assets/css/all.css`

giorgiosironi: Some sequencing of gulp tasks required. Apart from the wrong folder I'm copying, we need to execute `extract-assets` after any of the other watches has completed their callback.

giorgiosironi: Made a couple of bugfixes, and `extractAssets` should run whenever the interesting tasks are done now.

README.md
```diff
@@ -140,6 +140,7 @@ ENVIRONMENT=development docker-compose build assets
 To watch for changes, run:
 
 ```
+docker-compose build  # only necessary after switching branch or installing new dependencies
```
thewilkybarkid: Does this need `-f docker-compose.yml -f docker-compose.watch.yml`?

giorgiosironi: `docker-compose.watch.yml` does not build new images, but only decorates the existing ones with volumes/ports/commands; at the moment, it should not be needed for the build phase

thewilkybarkid: Something to remember when that's no longer true.

bin/watch
```diff
@@ -13,17 +14,18 @@ function cleanup {
 cleanup pattern-library-gulp-watch
 cleanup pattern-library-php-generate
 cleanup pattern-library-php-watch
-docker-compose build
```
thewilkybarkid: For speed?

README.md
```diff
@@ -151,16 +152,35 @@ You can pass options to the underlying gulp:
...
+To run additional gulp command in the same container where `gulp:watch` is running:
+
+```
+$ docker exec -it pattern-library-gulp-watch /bin/bash
+elife@...$ gulp test:unit
+```
+
+To extract the latest assets _while the watch is running_:
+```
+$ bin/extract-assets
```
thewilkybarkid: Doesn't exist?

thewilkybarkid: Was removed in e54806c79494afba9b1982838f8c662bfc8e7153.

thewilkybarkid:
Feels like `docker-compose.watch.yml` could be `docker-compose.override.yml` (ie for local development) and `bin/dev` could be removed.

giorgiosironi:
Implications:
- volumes will always be mounted, so `-f docker-compose.yml` will need to be specified in case we want to replicate a `ci` failure
- `bin/dev` removed means only supporting containers for local development rather than installing `node` on the host (I could remove it later to leave a fallback)
@davidcmoulton confirms?

bin/watch
giorgiosironi: Yes, even when everything is cached it would still add several seconds to the startup time of this script

giorgiosironi:
Updated documentation after removal of script, waiting on confirmation on previous points.

davidcmoulton:
Let's keep in `bin/dev`, so there's a fallback for now. Can be removed later once  local development using containers has proven stable.

giorgiosironi:
Remaining point is whether locally the containers should should always be setup for watching rather than using the code inside the last image that has been built (renaming `docker-compose.override.yml` so that it is picked up automatically).

davidcmoulton:
For ease of local development I'd go for having the containers always set up for watching.

PR #862, opened 2018-05-01
[WIP] ELPP-3458: Add new font loading pattern

davidcmoulton:
The intention is to allow fonts to be both preloaded and utilised efficiently.


source/_patterns/00-atoms/components/font-loader.mustache
```diff
@@ -0,0 +1,16 @@
+{{#fonts}}
+  <link rel="preload" href="{{#assetRewrite}}{{assetsPath}}{{{path}}}{{/assetRewrite}}" as="font">
+{{/fonts}}
+
+{{#fonts}}
+
+  <style>
+    @font-face {
+      {{#display}}font-display: {{display}};{{/display}}
```
thewilkybarkid: I don't think implements should have to know what these are, knowledge should remain here.

source/non-pattern-templates/for-head.mustache
```diff
@@ -0,0 +1,7 @@
+{{! These are the fonts specified in the the css's @font-face rules. }}
+
+<link rel="preload" href="{{#assetRewrite}}/fonts/249228f0-61ac-40cc-a5a5-5609c9816e3f.woff2{{/assetRewrite}}" as="font">
+
+<link rel="preload" href="{{#assetRewrite}}/fonts/627fbb5a-3bae-4cd9-b617-2f923e29d55e.woff2{{/assetRewrite}}" as="font">
```
thewilkybarkid: Include `type="font/woff2"`?

source/non-pattern-templates/for-head.mustache
```diff
@@ -0,0 +1,7 @@
+{{! These are the fonts specified in the the css's @font-face rules. }}
+
+<link rel="preload" href="{{#assetRewrite}}/fonts/249228f0-61ac-40cc-a5a5-5609c9816e3f.woff2{{/assetRewrite}}" as="font">
```
thewilkybarkid: Need to include `{{assetsPath}}` at the beginning:

https://github.com/elifesciences/pattern-library/blob/8b52fb3f4df69c2fbb94a5fd6bfe215fca834800/source/_patterns/00-atoms/components/orcid.mustache#L5

assets/sass/_fonts.scss
```diff
@@ -0,0 +1,16 @@
+// Each font referenced here should be preloaded in for-head.mustache
+
+/* Demi */
+@font-face {
+  font-display: fallback;
+  font-family: "Avenir Next Webfont";
+  src: url("../fonts/249228f0-61ac-40cc-a5a5-5609c9816e3f.woff2") format("woff2");
```
thewilkybarkid: Where did the font name come from? Since we're caching busting elsewhere, could be something easier to work with?

source/non-pattern-templates/for-head.mustache
davidcmoulton: Yes. It'll need `crossorigin="anonymous"`, too.

davidcmoulton:
Closing for now as other web fonts work takes priority. May reopen subsequently, if still needed.

davidcmoulton:
Reopening the PR now #865 is completed.

assets/sass/_fonts.scss
```diff
@@ -0,0 +1,29 @@
+// Should be preloaded in for-head.mustache
+@font-face {
+  font-display: fallback;
+  font-family: "Noto Sans";
+  src: url("../fonts/NotoSans-Regular-webfont-custom-2-subsetting.woff2") format("woff2");
```
davidcmoulton: @thewilkybarkid Will these paths (this one and the three more below on lines 12, 20 & 27) be okay for cache busting? I think we need to retain the information that exists in the filename, but can reformat it if that would make things easier.

thewilkybarkid: Should be fine (we do the same for images already).

thewilkybarkid:
Proposal: just have a `preload.json` file (contains path and type) that Journal can parse and use (this should refer to assets loaded from CSS/JS). Needs more work that end, but shouldn't be too difficult and means they can become headers.

davidcmoulton:
I really like that idea. Will throw something together.

davidcmoulton:
Closing in favour of https://github.com/elifesciences/pattern-library/pull/881/files

PR #865, opened 2018-05-14
ELPP-3626: Update web fonts

davidcmoulton:
Replaces previous web fonts with Noto Sans Regular, Noto Sans SemiBold, Noto Serif Regular and Noto Serif Bold.

Noto Serif Bold has been subsetted to basic latin only.

The other fonts have been subsetted to this glyph set, known internally as 'custom 2'

noto-sans-custom-2-subsetting-preview
which was generated using https://www.fontsquirrel.com/tools/webfont-generator with the settings
noto-sans-custom-2-subsetting-settings


giorgiosironi:
Passing comment: it would be useful to understand what text this change applies to, in order to understand if we have enough fixture data in the pattern-library website. For example, author names like `Tkáč` but also Greek letters `hL1ORF1p-ΔNTR` and other units of measurement `Å`.

assets/fonts/fonts.css
```diff
@@ -1,17 +1,61 @@
...
+  font-family: "Noto Sans";
+  src: url("NotoSans-Regular-webfont-western-latin.woff2") format("woff2");
+}
+
+/* Bold */
+@font-face {
+  font-display: fallback;
+  font-family: "Noto Sans";
+  src: url("NotoSans-Bold-webfont-western-latin.woff2") format("woff2");
+  font-weight: 700;
```
davidcmoulton: Consider whether `font-weight` (here and on line 24) should now be set to `bold` (or `800`), now we're not using a demi bold. (This would require corresponding changes in the patterns' CSS.)

davidcmoulton:
@giorgiosironi This change will apply to all text on the site. Part of Nick's review will be to ensure that these fonts have the range of glyphs required for representing our content corpus.

Good point about having sufficient representative content characters in the pattern-library sample content. That might be a separate pull request.

assets/fonts/fonts.css
davidcmoulton: Confirmed with design that all bold-like weights are now one.

To be expressed by `font-weight: bold;`

thewilkybarkid:
Dropping WOFF intentional?

davidcmoulton:
@thewilkybarkid Yes, woff2 has excellent browser support now: https://caniuse.com/#search=woff2. Those that can't use it will just use the fallback font.

thewilkybarkid:
Could sub-setting be part of the build?

davidcmoulton:
Resubsetted taking into account the unicode detective work of @thewilkybarkid. Unfortunately this has trebled the size of the overall font payload. We have had many complaints about missing glyphs in the current font set though, so this is probably worth doing anyway. We might be able to save some size by adding in specific code points rather than whole code blocks, but as content is being added all the time, it's likely that over time we'll need most of the glyphs in the subsets included in this pull request anyway.

davidcmoulton:
@thewilkybarkid Interesting idea. I originally thought we might use the CSS `unicode-range` so the browser only downloads the glyphs it needs, with the value of this property determined somewhere in the backend (and I guess maybe recalculated every ingest), but I'm not sure whether that plays nicely with preloading fonts though: might have to set up some tests to see what happens.

Subsetting as part of the build might work, not sure offhand what we could do to automate that, but worth looking into 🤔Would want to have a warning that triggers if the subsetted fonts grow over a certain collective size, to make sure that things don't go crazy.

thewilkybarkid:
What we really need to analyse is the percentage of pages that each character appears on (Google Analytics probably the quickest way to produce a list of URLs?). `A-Za-z0-9`, for example, will be on _a lot of_ pages, so should be part of the base set. A dollar sign, however, isn't, so would be in a separate (and reasonably small) subset ('misc ASCII'). The character-to-font mapping (which PL uses to generate the separate font files) could be shared with Journal, which works out which fonts a page needs and preloads them (caching the result, since it won't change very often (ie the text on the page changes); another thing that will complicate ESIs). Most of the work can be done by a shared (PHP) library.

We shouldn't actually automate the subset determination though, since a single character move from one subset to another will bust the cache (so a human should confirm the character-to-font mapping).

davidcmoulton:
That looks like an interesting approach. Want to get this PR soon though, so we can look at that as a separate piece of work.

PR #866, opened 2018-05-15
ELPP-3598 Remove non-ASCII characters from error pages

giorgiosironi:
Error pages and up in places like Fastly's VCL configuration files, where only ASCII characters are supported.

The right single quotation mark are outside of the 0x00-0x7F ASCII range. I'm no typographer but this character does not seem correct in this context. I've found no other non-ASCII characters being used (no emojis in error pages :pensive:).

The additional check quickly greps for any non-ASCII character in that folder. I've put it into `.ci` as that is a standard being adopted across projects to put checks to run in parallel.

In future use cases, non-ASCII characters can still be supported simply by escaping them as HTML entities: `&#8217;`|`&#x2019;`|`&rsquo;`.

### DIFF
--- a/.ci/errors-ascii-chars
+++ b/.ci/errors-ascii-chars
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -e
+
+echo "Checking for non-ASCII characters in error messages"
+echo "These messages end up in places with limitations like VCL"
+negated_ascii_range="[^\x00-\x7F]"
+matching=$(grep \
+    --color='auto' \
+    --perl-regexp "$negated_ascii_range" \
+    --line-number \
+    --recursive \
+    source/_patterns/00-atoms/errors/)
+if [ "$matching" != "" ]; then
+    echo "Non-ASCII characters found:"
+    echo "$matching"
+    exit 1
+fi
--- a/.ci/errors-ascii-chars
+++ b/.ci/errors-ascii-chars
@@ -15,3 +15,4 @@ if [ "$matching" != "" ]; then
     echo "$matching"
     exit 1
 fi
+exit 0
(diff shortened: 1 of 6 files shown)


.ci/errors-ascii-chars
```diff
@@ -0,0 +1,18 @@
...
+echo "These messages end up in places with limitations like VCL"
+negated_ascii_range="[^\x00-\x7F]"
+matching=$(grep \
+    --color='auto' \
+    --perl-regexp "$negated_ascii_range" \
+    --line-number \
+    --recursive \
+    source/_patterns/00-atoms/errors/ || true)
+if [ "$matching" != "" ]; then
+    echo "Non-ASCII characters found:"
```
davidcmoulton: Could this be expanded to state which file's contents generated has the error?

giorgiosironi: It does state this:
```
$ .ci/errors-ascii-chars 
Checking for non-ASCII characters in error messages
These messages end up in places with limitations like VCL
Non-ASCII characters found:
source/_patterns/00-atoms/errors/server-error.mustache:13:    <p>Something has gone seriously wrong.<br>We’ve logged the problem and will look into it.</p>
```

giorgiosironi: I have forced colors to be enabled:

.ci/errors-ascii-chars
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+
+echo "Checking for non-ASCII characters in error messages"
+echo "These messages end up in places with limitations like VCL"
```
thewilkybarkid: Feels like we should be isolating this to VCL, rather than placing limitations here.

thewilkybarkid: Hmm, just seen https://github.com/elifesciences/builder/pull/319#issuecomment-389158409.

.ci/errors-ascii-chars
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+
+echo "Checking for non-ASCII characters in error messages"
```
thewilkybarkid: `ISO-8859-1` (`ASCII` is a subset).

PR #868, opened 2018-05-21
ELPP-3614 Add form-field-info-link pattern & use as partial

davidcmoulton:


source/_patterns/00-atoms/forms/form-field-info-link.yaml
```diff
@@ -0,0 +1,21 @@
+assets:
+  css:
+    - form-field-info-link.css
+  js: []
+schema:
+  $schema: http://json-schema.org/draft-04/schema#
+  type: object
+  properties:
+    text:
```
thewilkybarkid: We've used `name` elsewhere.

assets/sass/patterns/atoms/form-field-info-link.scss
```diff
@@ -0,0 +1,15 @@
+@import "../../definitions";
+
+.form-field-info-link {
+  color: $color-text-secondary;
+  display: block;
```
thewilkybarkid: Causes the link to be full width.

PR #869, opened 2018-05-23
ELPP-3654: Cookie consent overlay

davidcmoulton:
Includes JavaScript/CSS changes only, this PR touches no patterns.

### DIFF
--- a/assets/js/components/CookieOverlay.js
+++ b/assets/js/components/CookieOverlay.js
@@ -13,6 +13,14 @@ module.exports = class CookieOverlay {
     this.window = _window;
     this.doc = doc;
 
+    try {
+      if (!window.elifeConfig.needsCookieConsent) {
+        return;
+      }
+    } catch (e) {
+      return;
+    }
+
     if (CookieOverlay.previouslyAccepted(this.doc)) {
       return;
     }
--- a/assets/js/components/CookieOverlay.js
+++ b/assets/js/components/CookieOverlay.js
@@ -36,7 +36,7 @@ module.exports = class CookieOverlay {
 
     $button.addEventListener('click', () => {
       this.$overlay.parentNode.removeChild(this.$overlay);
-      const expiryDate = 'Fri, 31 December 2038 23:59:59 GMT';
+      const expiryDate = 'Tue, 19 January 2038 03:14:07 UTC';
       this.doc.cookie = `cookieNotificationAccepted=true; expires=${expiryDate} path=/; domain=elifesciences.org;`;
     });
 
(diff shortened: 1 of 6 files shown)


assets/js/components/CookieOverlay.js
```diff
@@ -0,0 +1,46 @@
...
+    this.$elm = $elm;
+    this.window = _window;
+    this.doc = doc;
+
+    if (CookieOverlay.previouslyAccepted(this.doc)) {
+      return;
+    }
+
+    const message = 'This site uses cookies to deliver its services and analyse traffic. By using ' +
+                    'this site, you agree to its use of cookies. <a href="/privacy">Learn more</a>.';
```
thewilkybarkid: This should be absolute (`https://elifesciences.org/privacy`).

assets/js/components/CookieOverlay.js
```diff
@@ -0,0 +1,46 @@
...
+  buildDOM(message) {
+    const $root = utils.buildElement('div', ['opt-in-message-bar']);
+    utils.buildElement('div', ['opt-in-message-bar__message'], message, $root);
+
+    const $buttonWrapper = utils.buildElement('div', ['opt-in-message-bar__button_wrapper'], '', $root);
+    const $button = utils.buildElement('button', ['button', 'button--default', 'button--small'], 'Got it', $buttonWrapper);
+
+    $button.addEventListener('click', () => {
+      this.$overlay.parentNode.removeChild(this.$overlay);
+      const expiryDate = 'Fri, 31 December 2038 23:59:59 GMT';
```
thewilkybarkid: Might want to choose `Tue, 19 Jan 2038 03:14:07 UTC` to avoid any Int32 problems.

assets/js/components/CookieOverlay.js
thewilkybarkid: Link text isn't ideal...

assets/js/components/CookieOverlay.js
```diff
@@ -0,0 +1,46 @@
...
+      this.doc.cookie = `cookieNotificationAccepted=true; expires=${expiryDate} path=/; domain=elifesciences.org;`;
```
thewilkybarkid: Not going to work for some places (eg local development), but we should turn this off there anyway.

thewilkybarkid: Could lose the `domain` (`path` seems redundant too), but means having a popup for each app in each instance.

davidcmoulton: `path` is needed so that someone landing on e.g. https://elifesciences.org/articles/[id], clicking 'Got it' and getting the cookie to hide the overlay, will still have the cookie when on a page that's not within `/articles/`.

thewilkybarkid: Thought the default was the root for some reason, but sensibly it's not.

PR #870, opened 2018-05-24
ELPP-3656: Put js imports at top level

davidcmoulton:
Move the imports out of branching code.

Browser capability checks removed from here should now be implemented by whatever code loads this file (implemented in https://github.com/elifesciences/journal/pull/964).


source/_patterns/00-atoms/components/mustard-cutting-js.json
```diff
@@ -0,0 +1,3 @@
+{
+  "scriptLoader": "{% for javascript in javascripts %}var $script{{loop.index}} = document.createElement('script');$script{{loop.index}}.src = '{{ asset(javascript) }}';$body.appendChild($script{{loop.index}});{% endfor %}"
```
davidcmoulton: This contains both JavaScript and implementation-specific twig.

source/_patterns/00-atoms/components/mustard-cutting-js.mustache
```diff
@@ -0,0 +1,14 @@
+<script>
+  if (
+      window.localStorage &&
+      document.querySelector &&
+      window.addEventListener &&
+      !!(document.createElement('div')).dataset)
+  {
+    {{#scriptLoader}}
+      var $body = document.querySelector('body');
+      {{{scriptLoader}}}
```
thewilkybarkid: I think implementers should just give a list of scripts to load, rather than the actual JS.

thewilkybarkid: Maybe on `window.elifeConfig`? That way this loader can be a normal script rather than a pattern?

PR #873, opened 2018-06-08
ELPP-3685 Improve hidden content and Hypothesis

thewilkybarkid:
Hypothesis emits a [`scrolltorange` event](https://github.com/hypothesis/client/blob/f609e769780b4710316da1bcbe9397bf241bf05c/src/annotator/guest.coffee#L155-L165), which we can use to only open the collapsed section that contains the annotation, rather than having to open all collapsed sections when we can recognise Hypothesis being opened. (Hat tip to @JCCR.)

Note if an annotation spans multiple sections (unlikely due to the Hypothesis length limit), only the collapsed section containing the start of the annotation is opened.


thewilkybarkid:
Need to look at someone annotation a figure supplement, since that probably doesn't work at the moment.

thewilkybarkid:
So figure supplements and collapsed captions don't work (though they won't be working now either).

thewilkybarkid:
Annotating an author popups causes it to annotated the author details section instead, but they're the same content so it's ok.

thewilkybarkid:
Popups like the menu won't work, but unlikely to actually be annotated. Is it possible to make these non-annotatable?

thewilkybarkid:
Thinking we should have a standard way of 'revealing' a node, which would open, rotate etc all the parent nodes so that it's displayed.

thewilkybarkid:
Product requirement to continue opening all the sections.

thewilkybarkid:
Collapsed captions need reworking as the text is removed from the DOM rather than being `visuallyhidden`, so annotations end up as orphaned.

thewilkybarkid:
Created ELPP-3700 to refactor the collapsed captions.

PR #876, opened 2018-06-13
ELPP-3524 Make domain of CookieOverlay configurable

giorgiosironi:
Now onto adding a test.


giorgiosironi:
Tests don't pass, probably because of hostnames issues on the cookie in the environment where they run.

giorgiosironi:
Also it seems a build failure like this leaves the services up, causing problems to other builds - first issue to be solved.

giorgiosironi:
Test failure was reproducible locally with Docker, and simply turns our PhantomJS doesn't load the pages on an actual hostname (`file://`), hence the failure to read cookies set for `localhost` or similar. `localhost` was working for running `tests-watch` but not for `ci` or `gulp` usage.

assets/js/components/CookieOverlay.js
```diff
@@ -23,8 +24,16 @@ module.exports = class CookieOverlay {
...
-  static previouslyAccepted(doc) {
-    return !!utils.getCookieValue('cookieNotificationAccepted', doc.cookie);
+  previouslyAccepted() {
+    return !!utils.getCookieValue('cookieNotificationAccepted', this.doc.cookie);
+  }
+
+  accept() {
+    this.$overlay.parentNode.removeChild(this.$overlay);
+    const expiryDate = 'Tue, 19 January 2038 03:14:07 UTC';
+    const domain = this.$elm.dataset.domain ? this.$elm.dataset.domain : 'elifesciences.org';
```
thewilkybarkid: Could extract from `this.window.location.hostname` rather than making it configuration?

thewilkybarkid: Alternatively we have been setting configuration on `window.elifeConfig`.

giorgiosironi: @davidcmoulton in a recent stand-up was more favorable to "inject from the server-side" than to "detect with JavaScript"; can also change to `window.elifeConfig`

giorgiosironi: Solved in-person: `window.elifeConfig` is the way to go

giorgiosironi:
Updated to use `window.elifeConfig`.

assets/js/components/CookieOverlay.js
```diff
@@ -23,8 +24,16 @@ module.exports = class CookieOverlay {
...
-  static previouslyAccepted(doc) {
-    return !!utils.getCookieValue('cookieNotificationAccepted', doc.cookie);
+  previouslyAccepted() {
+    return !!utils.getCookieValue('cookieNotificationAccepted', this.doc.cookie);
+  }
+
+  accept() {
+    this.$overlay.parentNode.removeChild(this.$overlay);
+    const expiryDate = 'Tue, 19 January 2038 03:14:07 UTC';
+    const domain = this.window.elifeConfig.domain ? this.window.elifeConfig.domain : 'elifesciences.org';
```
davidcmoulton: As `elifeConfig` is set externally, access to its properties should be wrapped in a `try/catch` in case it's not available for some reason. Additionally, logging to New Relic could be helpful, for an example of this see the catch block at https://github.com/elifesciences/pattern-library/blob/e4e323666e56701d0f06437841dea275a4c8f333/assets/js/elife-loader.js#L25

giorgiosironi:
Created a utility function to deal with errors since there were a few other cases in `Hypothesis*` classes.

assets/js/components/CookieOverlay.js
```diff
@@ -12,8 +12,16 @@ module.exports = class CookieOverlay {
     this.$elm = $elm;
     this.window = _window;
     this.doc = doc;
+    try {
+      this.domain = this.window.elifeConfig.domain;
```
davidcmoulton: I think this should be:

```this.domain = this.window.elifeConfig.domain || 'elifesciences.org';``` 

That way, if the `domain` property is available, but has been set to an empty string, the default value will be properly assigned.

PR #877, opened 2018-06-13
Fix positioning of last nav-secondary__item

thewilkybarkid:
Currently:

screen shot 2018-06-13 at 16 12 00


thewilkybarkid:
Hmm, doesn't work in Firefox.

PR #879, opened 2018-06-14
Run smoke tests through common step

giorgiosironi:


docker-compose.ci.yml
```diff
@@ -0,0 +1,3 @@
+version: '3'
+
+services: {}
```
thewilkybarkid: Required by `dockerComposeProjectTestsParallel`. Should be a standard so no need to make it optional.

PR #881, opened 2018-06-15
ELPP-3458: Preloadable assets

davidcmoulton:
Adds a `preload.json` file to define assets for preloading.

Also moves `@font-face` declarations into scss.


assets/preload.json
```diff
@@ -0,0 +1,19 @@
+{
+  "resources": [
+    {
+      "type": "font/woff2",
```
thewilkybarkid: Include `as` too.

assets/preload.json
```diff
@@ -0,0 +1,19 @@
+{
```
thewilkybarkid: Can just be a list of objects?

assets/preload.json
```diff
@@ -0,0 +1,19 @@
...
+  "resources": [
+    {
+      "type": "font/woff2",
+      "paths": "fonts/NotoSans-Regular-webfont-custom-2-subsetting.woff2"
```
thewilkybarkid: `path`?

assets/preload.json
```diff
@@ -0,0 +1,19 @@
...
+      "type": "font/woff2",
+      "paths": "fonts/NotoSans-SemiBold-webfont-custom-2-subsetting.woff2"
+    },
+    {
+      "type": "font/woff2",
+      "paths": "fonts/NotoSerif-Regular-webfont-custom-2-subsetting.woff2"
+    },
+    {
+      "type": "font/woff2",
+      "paths": "fonts/NotoSerif-Bold-webfont-basic-latin-subsetting.woff2"
```
thewilkybarkid: #862 wasn't going to preload this.

assets/preload.json
davidcmoulton: Can the `as` value be reliably derived from the `type`?

assets/preload.json
davidcmoulton: Thanks, rushing.

assets/preload.json
davidcmoulton: Yup. Old habits worrying about JSON hijacking, will update.

assets/preload.json
thewilkybarkid: In the event we start including other things might become tricky. Not sure we would, but isn't onerous to just set it here.

assets/preload.json
thewilkybarkid: Even `uri`, for foreign resources?

thewilkybarkid: Maybe ignore for now.

thewilkybarkid:
Can remove `WIP`?

PR #882, opened 2018-06-18
Inline SocialLinks SVGs

thewilkybarkid:
Currently have to load an image, but the hover state is embedded. This follows SocialMediaSharers in embedded the SVG icons.

Means:

- 6 image requests less
- HTML is a fraction bigger (but should compress better)
- CSS is significantly smaller
- Browsers that don't support SVGs don't see anything (🤷‍♂️)


source/_patterns/01-molecules/components/social-links.mustache
```diff
@@ -1,23 +1,46 @@
-
 <div class="social-links">
   <ul class="social-links__list">
     <li class="social-links__list_item">
-      <a href="https://www.facebook.com/elifesciences" class="social-links__list_link social-links__list_link--facebook"><span class="visuallyhidden">Facebook</span></a>
+      <a href="https://www.facebook.com/elifesciences" class="social-links__list_link" aria-label="Facebook">
+        <svg width="20" height="20" aria-hidden="true">
```
thewilkybarkid: Is `aria-hidden` actually needed if it's inside `<a aria-label="...">`?

davidcmoulton: No, `aria-hidden` is not needed in this context.

PR #887, opened 2018-06-20
ELPP-3676: Update site header main navigation

davidcmoulton:
Removes the 'Labs' link in the secondary navigation, and adds an 'Innovation' link in the primary navigation.


assets/sass/patterns/molecules/site-header-nav-bar-secondary.scss
```diff
@@ -34,7 +34,7 @@
   }
 }
 
-.nav-secondary__item:nth-child(4).nav-secondary__item:nth-last-child(2) {
+.nav-secondary__item:nth-child(3).nav-secondary__item:nth-last-child(2) {
```
thewilkybarkid: Could we rethink this?

source/_patterns/00-atoms/components/opt-in-message-bar.json
```diff
@@ -0,0 +1,8 @@
+{
```
thewilkybarkid: Assume this isn't meant?

PR #888, opened 2018-06-21
Avoid exception inception

davidcmoulton:


davidcmoulton:
Realise I should have named this pull request "Exception inception prevention". Naming things is indeed hard.

PR #892, opened 2018-06-26
ELPP-3736: Remove individual css

davidcmoulton:
We don't and won't use the individual css files, so there's no point generating them.

This also means we can drop the requirement for Compass which is no longer supported, and will avoid the problem where triggering a (Compass-based) css file regeneration while the previous generation was still running caused the css-watch part of the watch process to hard stop, resulting in only a partial watch, and requiring a process restart to restore things.

The result of this PR being merged must be incorporated into https://github.com/elifesciences/patterns-php/pull/637

### DIFF
--- a/Dockerfile.assets-builder
+++ b/Dockerfile.assets-builder
@@ -20,15 +20,11 @@ RUN npm install -g gulp-cli
 RUN apt-get update && apt-get install -y \
     bzip2 \
     g++ \
-    make \
-    ruby \
-    ruby-dev 
-
-RUN gem install compass
+    make
 
 USER elife
 ENV PROJECT_FOLDER=/srv/pattern-library
-# for simplicity, also set right permissions to test folder 
+# for simplicity, also set right permissions to test folder
 # (and any volume mounted on it) in case assets-builder
 # is used locally to run gulp tasks
 RUN mkdir ${PROJECT_FOLDER} && \
@@ -46,7 +42,6 @@ COPY --chown=elife:elife \
     .jscsrc \
     .jshintrc \
     .stylelintrc \
-    config.rb \
     gulpfile.js \
     ${PROJECT_FOLDER}/
 
@@ -61,4 +56,4 @@ RUN gulp --environment ${environment} js
 COPY --chown=elife:elife source/ ${PROJECT_FOLDER}/source
 
 COPY --chown=elife:elife assets/sass ${PROJECT_FOLDER}/assets/sass
-RUN gulp --environment ${environment} generateStyles
+RUN gulp --environment ${environment} generateCss
--- a/bin/validate
+++ b/bin/validate
@@ -23,17 +23,11 @@ $jsonSchema = $refResolver->resolve('file://' . __DIR__ . '/../extras/json-schem
 $exit = 0;
 
 foreach ($finder as $file) {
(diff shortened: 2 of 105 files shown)


Dockerfile.assets-builder
```diff
@@ -61,4 +56,4 @@ RUN gulp --environment ${environment} js
 COPY --chown=elife:elife source/ ${PROJECT_FOLDER}/source
 
 COPY --chown=elife:elife assets/sass ${PROJECT_FOLDER}/assets/sass
-RUN gulp --environment ${environment} generateStyles
+RUN gulp --environment ${environment} generateCss
```
giorgiosironi: this is the only external reference to this gulp task

Dockerfile.assets-builder
```diff
@@ -20,15 +20,11 @@ RUN npm install -g gulp-cli
 RUN apt-get update && apt-get install -y \
```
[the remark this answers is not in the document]
thewilkybarkid: Should be able to update the base image now.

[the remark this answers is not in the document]
giorgiosironi: https://github.com/elifesciences/pattern-library/pull/892/files#diff-e6b04132365480be9a1e78d6cc8987b5R1 can be updated too without the dependency on Ruby, but definitely not now, best done as a follow-up

Dockerfile.assets-builder
thewilkybarkid: We shouldn't globally install Gulp, and instead use the local version.

thewilkybarkid: Yes, can follow up with a tidy up.

davidcmoulton: I've renamed that gulp task to `generateCss` it as it's doing a slightly different job now. It runs as part of the overall gulp watch.

Dockerfile.assets-builder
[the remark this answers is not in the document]
thewilkybarkid: Going to create a separate tidy-up PR.

bin/validate
```diff
@@ -36,7 +36,7 @@ foreach ($finder as $file) {
     }
 
     $validator->reset();
-    $validator->check($yaml->schema, $jsonSchema);
+    $validator->check($yaml, $jsonSchema);
```
davidcmoulton: Not sure this & line 60 is right. Script exits with 0, but with the message `Json Pointer '/schema' references a nonexistent value`.

thewilkybarkid: https://github.com/elifesciences/pattern-library/blob/3f04e2b1f0c5e2a3d267368c48866765b6e83945/source/_patterns/00-atoms/components/author-details.yaml#L38 (and similar) need to lose their pointers (=fragments), since it doesn't exist anymore.

davidcmoulton: Running validate now on: https://github.com/elifesciences/pattern-library/pull/892/files#diff-460b5554de13b09ccc95e53925083a87R38 now gives `Pointer is not a string` (exit code 0).

thewilkybarkid: Fixed in 77d75aa3835769bb23698f523914a303aab1596f.

thewilkybarkid: (Though need to look at the exit code.)

PR #893, opened 2018-06-26
Tidy containers

thewilkybarkid:
Follow up to (and so currently includes) #892.

### DIFF
--- a/.env
+++ b/.env
@@ -3,5 +3,6 @@ IMAGE_TAG=develop
 ENVIRONMENT=production
 DESCRIPTION=
 TEST_HTML=
+NODE_VERSION=6.14.3
 # add '-debug' for VNC support
 SELENIUM_IMAGE_SUFFIX=
--- a/Dockerfile.assets
+++ b/Dockerfile.assets
@@ -2,11 +2,12 @@ ARG image_tag=latest
 FROM elifesciences/pattern-library_assets-builder:${image_tag} AS assets-builder
 
 FROM tianon/true:latest
-ENV PROJECT_FOLDER=/srv/pattern-library
-WORKDIR ${PROJECT_FOLDER}
 
-COPY --from=assets-builder /srv/pattern-library/source/ ${PROJECT_FOLDER}/source
-COPY assets/preload.json ${PROJECT_FOLDER}/source/assets
+WORKDIR /srv/pattern-library
+
+COPY source/ source/
+COPY --from=assets-builder /source/assets/ source/assets/
+COPY assets/preload.json source/assets/
 
 ARG description=
 LABEL description=${description}
--- a/Dockerfile.assets
+++ b/Dockerfile.assets
@@ -1,6 +1,5 @@
 ARG image_tag=latest
 FROM elifesciences/pattern-library_assets-builder:${image_tag} AS assets-builder
-
 FROM tianon/true:latest
 
 WORKDIR /srv/pattern-library
--- a/Dockerfile.assets
+++ b/Dockerfile.assets
@@ -5,7 +5,7 @@ FROM tianon/true:latest
 WORKDIR /srv/pattern-library
 
 COPY source/ source/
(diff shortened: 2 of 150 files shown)


.env
```diff
@@ -3,5 +3,6 @@ IMAGE_TAG=develop
 ENVIRONMENT=production
 DESCRIPTION=
 TEST_HTML=
+NODE_VERSION=6.14.3
```
davidcmoulton: This makes my eyes bleed. Update of this to LTS coming soon (as a separate PR).

giorgiosironi: Expires in 2019 fortunately

Dockerfile.assets
```diff
@@ -1,12 +1,12 @@
 ARG image_tag=latest
 FROM elifesciences/pattern-library_assets-builder:${image_tag} AS assets-builder
-
 FROM tianon/true:latest
-ENV PROJECT_FOLDER=/srv/pattern-library
```
giorgiosironi: `PROJECT_FOLDER` is sometimes read by pipelines to determine where to find the smoke tests. This image should not be affected https://github.com/elifesciences/elife-jenkins-workflow-libs/search?q=PROJECT_FOLDER&unscoped_q=PROJECT_FOLDER

Dockerfile.assets-builder
```diff
@@ -1,59 +1,27 @@
...
-# end base image
-
-# potentially base image
-RUN npm install -g gulp-cli
-# end potentially base image
-
-RUN apt-get update && apt-get install -y \
-    bzip2 \
-    g++ \
-    make
```
giorgiosironi: Where have `bzip2`, `g++` and `make` gone? Not necessary anymore, what they were used for?

Dockerfile.assets-builder
```diff
@@ -1,59 +1,27 @@
...
-COPY --chown=elife:elife assets/fonts/ ${PROJECT_FOLDER}/assets/fonts
-RUN gulp --environment ${environment} fonts
-COPY --chown=elife:elife assets/img/ ${PROJECT_FOLDER}/assets/img
-RUN gulp --environment ${environment} img
+COPY assets/fonts/ assets/fonts/
+RUN node_modules/.bin/gulp --environment ${environment} fonts
 
-COPY --chown=elife:elife assets/js/ ${PROJECT_FOLDER}/assets/js
-RUN gulp --environment ${environment} js
+COPY assets/img/ assets/img/
```
giorgiosironi: Containers running as `root` in general should be avoided to give them the least privilege. This is a build phase so may be different. Then the `node` image uses the `root` user as default :thinking:

docker-compose.override.yml
```diff
@@ -1,13 +1,19 @@
...
+            - ./package.json:/package.json
+            - node_modules:/node_modules
     assets-builder:
         volumes:
             - ./:/srv/pattern-library
             - /srv/pattern-library/node_modules
             - source_assets:/srv/pattern-library/source/assets
             - ./.container_source_assets:/srv/pattern-library/.container_source_assets
             - test_build:/srv/pattern-library/test/build
+            - node_modules:/node_modules
```
giorgiosironi: there are two `node_modules` folders in this list now?

Dockerfile.assets-builder
thewilkybarkid: The `node` image extends `buildpack-deps` which already contains them.

Dockerfile.assets-builder
thewilkybarkid: Done in 769c7aa

thewilkybarkid:
@davidcmoulton Should be fixed.

thewilkybarkid:
The slow behaviour was happening to me before this change, are you only now seeing it?

davidcmoulton:
It seems worse now. If you've seen it before though then doesn't seem like it's introduced by this change, so can ignore for this PR.

PR #903, opened 2018-07-26
Support captioned iframe assets

nlisgo:
Improves: https://github.com/elifesciences/jira-import/issues/3555


thewilkybarkid:
Looks good, we just need to lose the bottom margin.

PR #906, opened 2018-09-06
Added aria label to Investor Logo's for accessibility.

discodavey:
resolves https://github.com/elifesciences/issues/issues/1478


source/_patterns/01-molecules/components/investor-logos.mustache
```diff
@@ -1,4 +1,4 @@
-<ol class="investor-logos">
+<ol class="investor-logos" role="contentinfo" aria-label="eLife is funded by these organisations">
```
davidcmoulton: @discodavey Good call with `role`.

PR #909, opened 2018-09-12
Add aria-label to Social Media in footer for Accessibility

discodavey:


source/_patterns/01-molecules/components/social-links.mustache
```diff
@@ -1,4 +1,4 @@
-<div class="social-links">
+<div class="social-links" role="contentinfo" aria-label="Social media links for eLife Sciences">
```
davidcmoulton: I wonder if the`contentinfo` is redundant here, as the footer is already the contentinfo in this context: https://w3c.github.io/aria-practices/examples/landmarks/contentinfo.html#tabpanel2 (The `social-links` pattern is only ever in the site footer.)

discodavey: I see what you mean and agree but as usual there are developers out there who think you should still add it in.

I will remove the role and leave the label if you're happy with that?

thewilkybarkid: Could use a comment to clarify why it’s not there?

source/_patterns/01-molecules/components/social-links.mustache
```diff
@@ -1,3 +1,4 @@
+<!-- No aria role contentinfo needed in Social Links as this is always placed within the footer -->
```
thewilkybarkid: Use Mustache comments (`{{! ... }}`), as we don't actually want this appearing in the real HTML.

PR #911, opened 2018-09-20
Add aria label for MSA's

discodavey:


source/_patterns/01-molecules/teasers/teaser.mustache
```diff
@@ -3,7 +3,7 @@
   {{! These are usually links to major subject areas, see teaser-main-article, but could be unlinked
       text expressing a relationship, e.g. 'Built upon by' in teaser-secondary-related-item}}
   {{#contextLabel}}
-    <ol class="teaser__context_label_list">
+    <ol class="teaser__context_label_list" aria-label="These subject areas are for the following article">
```
thewilkybarkid: eLife is confused about its terminology. Half say 'subject', half say 'research category'! (Hence the different in the URI and title on https://elifesciences.org/subjects)

discodavey: Andy has confirmed my label is good to go. So just needs to be approved.

PR #912, opened 2018-09-20
Add aria label for list of authors

discodavey:


source/_patterns/02-organisms/content-headers/content-header.mustache
```diff
@@ -118,7 +118,7 @@
 
   {{#authors}}
     <div class="content-header__authors">
-      <ol class="content-header__author_list">
+      <ol class="content-header__author_list"  aria-label="List of authors of this article">
```
thewilkybarkid: Is `List of` needed?

PR #914, opened 2018-09-20
aria label for ORCID ID's

discodavey:


source/_patterns/00-atoms/components/orcid.mustache
```diff
@@ -1,4 +1,4 @@
-<span class="orcid">
+<span class="orcid" aria-label="An ORCID iD is a persistent digital identifier for researchers">
```
thewilkybarkid: Doesn't read like a label, but a description of what ORCID is...

davidcmoulton: Doesn't work in this context. Would be appropriate to expand the spoken description for the login button though, which means in `journal` adding to:
- https://github.com/elifesciences/journal/blob/105132a10587eaf3b9f0b7d796a9766c0ba7d52c/test/Controller/AuthenticationTest.php#L29
- https://github.com/elifesciences/journal/blob/105132a10587eaf3b9f0b7d796a9766c0ba7d52c/test/Controller/AuthenticationTest.php#L149
and
- https://github.com/elifesciences/journal/blob/fd91c74ba4e4f060c3bfda759dc5402f6f5f1749/src/ViewModel/Factory/SiteHeaderFactory.php#L159

Ensure to update the json files for the [pattern library example pages](https://github.com/elifesciences/pattern-library/tree/develop/source/_patterns/04-pages) too, to reflect the updated text.

Still need to work out the working to announce the ORCID in the articles.

discodavey:
Original text that ORCID have supplied is now as agreed in the login/register link at the top of the page. 

The ORCID link within the author information section has been reduced from two links to one as both had a different destination. The one link now only goes to the author ID. I have used the visually hidden class as an aria-label replaces the link text.

source/_patterns/00-atoms/components/orcid.mustache
```diff
@@ -1,10 +1,9 @@
...
+  <a href="https://orcid.org/{{id}}">
     <picture>
       <source srcset="{{#assetRewrite}}{{assetsPath}}/img/icons/orcid.svg{{/assetRewrite}}" type="image/svg+xml">
       <img src="{{#assetRewrite}}{{assetsPath}}/img/icons/orcid.png{{/assetRewrite}}" class="orcid__icon"
            alt="ORCID icon">
-    </picture>
-  </a>
-  <a href="https://orcid.org/{{id}}" class="orcid__id">{{id}}</a>
-</span>
+    </picture> <span class="visuallyhidden">"This ORCID iD identifies the author of this article:"</span>
```
davidcmoulton: Don't need the extra `iD` as the "ID" in `ORCID` has it covered.

source/_patterns/00-atoms/components/orcid.mustache
```diff
@@ -1,10 +1,9 @@
...
+	 {{id}}</a>
+</span>
```
davidcmoulton: All files should end with one line break.

source/_patterns/04-pages/article--research.json
```diff
@@ -529,7 +529,7 @@
...
...ds was controlled during training and experimental periods to promote behavioral motivation. All surgical procedures were performed under anesthesia, and all efforts were made to minimize post-surgical pain or suffering. Institutional veterinarians continually monitored animal health and well-being.</p>    </div>  </section>  <section class=\"article-section \">    <header class=\"article-section__header\">      <h3 class=\"article-section__header_text\">Reviewers</h3>    </header>    <div class=\"article-section__body\">      <ol class=\"list\">        <li>Sabine Kastner, Reviewing editor, Princeton University, United States</li>      </ol>    </div>  </section>  <section class=\"article-section \">    <header class=\"article-section__header\">      <h3 class=\"article-section__header_text\">Copyright</h3>    </header>    <div class=\"article-section__body\">      <p>© 2016, Schaffelhofer et al.</p><p>This article is distributed under the terms of the <a href=\"http://creativecommons.org/licenses/by/4.0/\">Creative Commons Attribution License</a>, which permits unrestricted use and redistribution provided that the original author and source are credited.</p>    </div>  </section>"
```
davidcmoulton: Will need to match changes to the pattern as appropriate (remove "iD").

source/_patterns/00-atoms/components/orcid.mustache
davidcmoulton: Ignore this comment. It appears ORCID use this tautology themselves, so we should too. Sorry for the confusion.

source/_patterns/04-pages/article--research.json
davidcmoulton: Ignore this comment, see above.

PR #916, opened 2018-09-25
Digest spacing and sample

davidcmoulton:
- Removes 48px margin from the top of a captioned asset when it's the first item in the `.content-container` (requested for Digests pages).
- Adds sample Digest page to pattern library

### DIFF
--- a/source/_patterns/04-pages/digest.json
+++ b/source/_patterns/04-pages/digest.json
@@ -1,5 +1,5 @@
 {
-  "isPrimaryResearch": true,
+  "isPrimaryResearch": false,
   "hasRelated": true,
   "siteHeader": {
     "homePagePath": "#",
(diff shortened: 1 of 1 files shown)


assets/sass/patterns/molecules/captioned-asset.scss
```diff
@@ -6,6 +6,12 @@
   @include blg-spacing("top", "medium", "margin");
   @include blg-spacing("bottom", "medium", "margin");
 
+  // Required for Digests when captioned-asset may be first piece of main content in the page. Prevents the top edge
+  // of the image frame sitting below any content appearing in the right hand column.
```
thewilkybarkid: Are there other cases where top margins are used?

davidcmoulton: There are a handful of cases: read more listings, inline asset viewer, profiles in the About section, the body of an article section when js show/hide is enabled and within the guts of some tables and content headers. It's pretty rare as the general rule is to apply space to the bottom of a component to space it from its following UI neighbour.

thewilkybarkid: I agree, but was wondering if this rule should be applied globally.

davidcmoulton: I considered that, but I think this is a safer option. If needs be we can expand the approach to other things that come in at the top of a Digest, but I doubt that'll be necessary.

PR #917, opened 2018-10-02
Add aria label for contextual data

discodavey:


source/_patterns/01-molecules/components/contextual-data.mustache
```diff
@@ -1,4 +1,4 @@
-<div class="contextual-data">
+<div class="contextual-data" aria-label="The following contains the number of views, cites and annotations in this article">
```
davidcmoulton: @discodavey I can't get this to read out in Chrome Vox. Does it read out on other screen readers you've tested? Wondering if it might be better on the [`ul.contextual-data__list`](https://github.com/elifesciences/pattern-library/pull/917/files#diff-1ea9590b56d1d6ac88b61bbd258462e1L4) element. What do you think?

source/_patterns/01-molecules/components/contextual-data.mustache
```diff
@@ -1,7 +1,7 @@
 <div class="contextual-data">
 
   {{#metricsData}}
-    <ul class="contextual-data__list">
+    <ul class="contextual-data__list" aria-label="The following contains the number of views, cites and annotations in this article">
```
davidcmoulton: Change "cites" to "citations" (well understood domain language).

PR #920, opened 2018-10-04
Bug Fix for Reset Button

discodavey:
This is to make the reset button show and hide when text is added or removed


discodavey:
Button will need to be styled as well

discodavey:
@davidcmoulton Styled reset button. Hidden text so screen reader can still read.

PR #925, opened 2018-10-16
Increase full-width teaser image breakpoint

thewilkybarkid:
Should fix https://github.com/elifesciences/issues/issues/4492#issuecomment-430197842.


thewilkybarkid:
Hold off for now, sounds like the spec might be changing.

thewilkybarkid:
It's not, this should now match #4492.

PR #926, opened 2018-10-19
Aria main menu accessibility

discodavey:


assets/js/components/MainMenu.js
```diff
@@ -25,9 +26,9 @@ module.exports = class MainMenu {
    */
   moveWithinDom() {
     this.$elm.classList.add('main-menu--js');
-    let $globalWrapper = this.doc.querySelector('.global-wrapper');
+    let $globalWrapper = this.doc.querySelector('.wrapper--site-header');
```
davidcmoulton: `$globalWrapper` should be renamed to something more appropriate now it's not referencing the .`global-wrapper` element.

As this line is being changed anyway, recommend using `const` in preference to `let`, as its value never changes.

discodavey:
Change done

PR #927, opened 2018-10-30
Remove last child as bug in firefox

discodavey:


discodavey:
@davidcmoulton Have removed the last-child as causing the issue. Tested Chrome,Firefox Edge and IE11 and in various responsive widths and no issues.

discodavey:
Tested Firefox versions 60-64 and all ok.

thewilkybarkid:
Might want to look at #877.

discodavey:
What was causing the issue in the screen grab on #877? I have removed the code and tested the browsers and not getting what you had. When did this occur as I cannot seem to replicate what you had in that ticket?

source/_patterns/01-molecules/navigation/login-control.mustache
```diff
@@ -1,5 +1,5 @@
 
-<div class="login-control"
+<div class="login-control {{#isLoggedIn}}login-control--logged-in{{/isLoggedIn}}"
```
davidcmoulton: Is this still needed?

assets/sass/patterns/molecules/site-header-nav-bar-secondary.scss
```diff
@@ -29,8 +29,8 @@
   @include padding(0 12 0 0);
   text-transform: uppercase;
 
-  &:last-child {
-    align-self: self-end;
+  &.nav-secondary__item--logged-in {
+    align-self: stretch;
```
davidcmoulton: This should only apply when JavaScript is available, we still want `baseline` cross axis alignment for the text displayed without JavaScript. Suggest the selector is modified to something like:

```.js &.nav-secondary__item--logged-in```

PR #928, opened 2018-11-06
Navigation connection detect

discodavey:


assets/js/elife-loader.js
```diff
@@ -1,15 +1,33 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  try {
+    var scriptPaths,
+        $body;
+    if (browserHasMinimumFeatureSupport() && !networkIsDefinitelySlow()) {
       scriptPaths = window.elifeConfig.scriptPaths;
```
davidcmoulton: This is what's currently breaking the tests, as `window.elifeConfig` is `undefined` at this point, and trying to reference a property of `undefined` is an error.

Could avoid this by passing the config object as an argument to `init()` (see comment for line 27 re `init`), so instead of lines 28 and 31 within `init`, something like:

```
function init(config) {
  var scriptPaths = config.scriptPaths;
  ...
}
```

and to invoke it:

```
init(window.elifeConfig || {});
```

The parametrised `init` function could be a good candidate for more tests, too, at least in order to verify that `scriptPaths` can be derived correctly when it's available, and that its absence doesn't break things.

test/elife-loader.spec.js
```diff
@@ -0,0 +1,92 @@
...
+const expect = chai.expect;
+const spy = sinon.spy;
+
+// load in component(s) to be tested
+let loader = require('../assets/js/elife-loader');
+
+describe('The eLife Loader', function () {
+  "use strict";
+
+  let loader;
```
davidcmoulton: Doesn't look to be used anywhere, delete.

test/elife-loader.spec.js
```diff
@@ -0,0 +1,92 @@
+const chai = require('chai');
+const sinon = require('sinon');
+
+const expect = chai.expect;
+const spy = sinon.spy;
+
+// load in component(s) to be tested
+let loader = require('../assets/js/elife-loader');
```
davidcmoulton: Is this needed? Delete if not. The load looks to be handled by https://github.com/elifesciences/pattern-library/pull/928/files#diff-4b3d23034489717c53e6b44e65e7fdc0R23.

test/elife-loader.spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+describe('The eLife Loader', function () {
+  "use strict";
+
+  let loader;
+
+  describe("checks if the network information", function() {
```
davidcmoulton: `describe` ideally should describe a thing, for example a function. So here it might be clearer to say:

```
describe('the isNetworkInformationAvailable() function', function () {
```

and then the tests within could be described in terms of:

```
  it('returns true when the network information is available'...);
  it('returns false when the network information is not available'...);
```

The advantage of this type of wording is that when a test fails, it's immediately clear which function has the problem, and which of its behaviours is breaking.

test/elife-loader.spec.js
```diff
@@ -0,0 +1,92 @@
...
+          return false;
+        };
+        expect(networkIsDefinitelySlow(isNetworkInfoAvailableMock)).to.be.false;
+      });
+      
+    });
+
+    context('when network information is available', function(){
+
+      context('network speed is 2g', function() {
```
davidcmoulton: Picky I know, but this might be slightly clearer:

```
context('the network speed is 2g', function() {
```

test/elife-loader.spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+          const navigatorConnectionMock = {
+            effectiveType: '2g'
+          }
+
+          expect(networkIsDefinitelySlow(isNetworkInfoAvailableMock,navigatorConnectionMock)).to.be.true;
+        });
+      });
+
+      context('network speed is greater than 2g', function() {
```
davidcmoulton:
```
context('the network speed is greater than 2g', function() {
```

test/elife-loader.spec.js
```diff
@@ -0,0 +1,92 @@
...
+            return true
+          };
+
+          const connectionSpeeds = ['3g', '4g'];
+          connectionSpeeds.forEach((speed) => {
+            const navigatorConnectionMock = {
+            effectiveType: speed
+          };
+
+          expect(networkIsDefinitelySlow(isNetworkInfoAvailableMock,navigatorConnectionMock)).to.be.false;
```
davidcmoulton: Missing a space after the comma.

assets/js/elife-loader.js
```diff
@@ -1,15 +1,33 @@
 (function (window) {
```
davidcmoulton: Don't run this through an IIFE, as could make testing more difficult.

assets/js/main.js
```diff
@@ -41,6 +41,7 @@ Components.ViewerModal = require('./components/ViewerModal');
 Components.ViewSelector = require('./components/ViewSelector');
 
 // App
+
```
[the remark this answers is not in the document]
discodavey: Completed

assets/js/elife-loader.js
```diff
@@ -1,15 +1,33 @@
...
+    if (!isNetworkInformationAvailable(navigator.connection)) {
+      return false;
+    }
+    if (navigator.connection.effectiveType.indexOf('2g') > -1) {
+      return true;
+    }
+    return false;
+  }
+
+  try {
```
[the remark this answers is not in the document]
discodavey: Completed

assets/js/elife-loader.js
discodavey: Completed

test/elife-loader.spec.js
discodavey: Completed

test/elife-loader.spec.js
discodavey: Completed

test/elife-loader.spec.js
discodavey: Completed

test/elife-loader.spec.js
discodavey: Completed

test/elife-loader.spec.js
discodavey: Completed

test/elife-loader.spec.js
discodavey: Completed

assets/js/elife-loader.js
discodavey: Completed

PR #930, opened 2018-11-27
Add aria-expanded, hidden and focus

discodavey:


assets/js/components/Popup.js
```diff
@@ -74,6 +74,7 @@ module.exports = class Popup {
 
     const closestPopup = utils.closest(e.target, '.popup');
     if (closestPopup && closestPopup === this.popupHitBox) {
+      this.popupHitBox.setAttribute('aria-hidden', true);
```
davidcmoulton: When the user clicks on the currently displayed popup, line 77 has the effect of setting `aria-hidden="true" on it, but doing nothing else. Is this the desired behaviour?

discodavey:
Now works on one click

assets/js/components/Popup.js
```diff
@@ -319,7 +324,22 @@ module.exports = class Popup {
...
       this.positionPopupHitBox(e);
     } else {
+      this.popupHitBox.setAttribute('aria-expanded', 'false');
+      this.popupHitBox.removeAttribute('tabindex', -1);
+      this.popupHitBox.setAttribute('aria-hidden', 'true');
       this.popupHitBox.style.display = 'none';
+      this.$link.focus();
+    }
+
+    if (this.isOpen) {
```
davidcmoulton: This should be combined with the identical conditional on [line 324](https://github.com/elifesciences/pattern-library/pull/930/files#diff-752da1b00c0db6cd94aafc27aa81c946R324).

assets/js/components/Popup.js
```diff
@@ -319,7 +324,22 @@ module.exports = class Popup {
...
+      this.popupHitBox.addEventListener('blur', () => {
+        this.popupHitBox.setAttribute('aria-expanded', 'false');
```
davidcmoulton: There's a lot of duplication here with the `else` block defined on [line 326](https://github.com/elifesciences/pattern-library/pull/930/files#diff-752da1b00c0db6cd94aafc27aa81c946R326). Suggest extracting the duplication into a separate method.

assets/js/components/Popup.js
discodavey: The reason why I moved this outside was it was causing the triple click

assets/js/components/Popup.js
discodavey: agree

assets/js/components/Popup.js
discodavey: Resolved as did not work on Tuesday hence reason why I separated but now does since method created. Beginning to think my Mac is playing up on me.

assets/js/components/Popup.js
discodavey: Completed

assets/js/components/Popup.js
```diff
@@ -260,8 +260,20 @@ module.exports = class Popup {
...
+    this.$link.parentNode.insertBefore(this.popupHitBox, this.$link);
     this.popupHitBox.style.left = `${left}px`;
     this.popupHitBox.style.top = `${top}px`;
+    this.popupHitBox.setAttribute('aria-expanded', 'true');
+    this.popupHitBox.setAttribute('tabindex', -1);
+    this.popupHitBox.setAttribute('aria-hidden', 'false');
+    this.popupHitBox.focus();
+  }
+
+  accessibilityPopupHitBox() {
```
davidcmoulton: Could probably do with a more descriptive name for this method.

Refactor suggestion: with a better name you could combine the code in this method with the complementary code on [lines 266](https://github.com/elifesciences/pattern-library/pull/930/files#diff-752da1b00c0db6cd94aafc27aa81c946R266) - 268

For example, something like:
```
setPopupHitBoxAttributes(isToBeShown) {
  if (isToBeShown) {
    // do the showing stuff
  } else {
    // do the hiding stuff
  }
}
```
and then call the same function in the places where you need to change the attributes (equivalent to lines 266 etc and lines 340) with different arguments .

Bearing in mind the legacy sprawl in this class, I think this helps organise it that little bit more without costing too much effort.

assets/js/components/Popup.js
```diff
@@ -318,8 +342,16 @@ module.exports = class Popup {
     // Changing the state depending on the other properties of the object.
     if (this.isOpen) {
       this.positionPopupHitBox(e);
+
+      this.popupHitBox.addEventListener('blur', () => {
```
davidcmoulton: This `blur` event listener callback  could ideally `removeEventListener` itself to prevent multiple event listener registrations when the same popup hit box is opened/closed/opened multiple times on a page. In reality I can't see that happening more than a handful of times, so it shouldn't present any performance issues. Because of this extremely low risk, I don't think any action is required.

PR #932, opened 2018-12-04
Remove event-stream

diversemix:
This is to remove the vulnerability of the malicious `event-stream` (3.3.5 and above). I have not updated `npm-shrinkwrap.json` as there were a lot of changes on my machine. Chances are that this could do will a closer look - however this should satisfy the vulnerability issue.

/cc @davidcmoulton @discodavey


giorgiosironi:
https://github.com/elifesciences/pattern-library/blob/develop/npm-shrinkwrap.json#L2306-L2310 seems to be already pinning `3.3.4` so it should be unchanged in practice, but its `from` needs updating. I'll try to do this with Docker.

giorgiosironi:
Ah I am on 4G at the moment, too slow for Docker usage, leaving it for later unless someone gets around to it first.

diversemix:
@giorgiosironi Sounds good - my understanding of `npm-shrinkwrap.json` was that it was allowing:
```
"event-stream@>=3.3.2 <4.0.0"
```

thewilkybarkid:
Is it currently pinned. It's only used by 1 line in our `gulpfile` so I'll look at removing it entirely.

diversemix:
@thewilkybarkid - I had a quick look at doing this - its only used in one place. I'm tied up today so go for it :+1:

diversemix:
Looks good :+1: I can't approve this (as author) - I'd have probably used the `await` syntax but cat skinning comes in multiple ways!

thewilkybarkid:
So would I, but it's still Node 6. Doesn't seem to be working properly yet.

diversemix:
Don';t know why the tests are failing ... they work for me:
```
npx gulp test:unit
[10:55:21] Using gulpfile ~/dev/pattern-library/gulpfile.js
[10:55:21] Starting 'browserify-tests'...
✔
```

thewilkybarkid:
I see `Finished 'browserify-tests' after 58 ms`, so it's not waiting for it to finish. Something inside the promise but not be promisified.

gulpfile.js
```diff
@@ -213,24 +212,25 @@ gulp.task('js:cs', () => {
...
+                      .bundle()
+                      .pipe(source(entry))
+                      .pipe(rename({dirname: ''}))
+                      .pipe(gulp.dest('./test/build'))
+                      .pipe(reload())
+                      .on('error', reject)
+                      .on('finish', resolve);
+              });
+          }));
+      });
```
davidcmoulton: Needs a `.catch`?

giorgiosironi:
Is this missing some work? To try https://github.com/elifesciences/issues/issues/4582 here we would need to update all packages first to avoid a flood of PRs, so if this is boing to be removed it's best to go through with it?

lsh-0:
we're still getting daily security alerts for pattern-library. 

just saying.

(and silently judging).

lsh-0:
hey hey! first day without a github security alert. thanks guys

PR #933, opened 2018-12-19
Remove width to make image 100% width

discodavey:


assets/sass/patterns/molecules/teaser.scss
```diff
@@ -186,16 +186,6 @@
 
 }
 
-@media only all and (min-width: #{$bkpt-site--medium}px) {
-
-  .teaser__picture {
```
davidcmoulton: Removing this block will affect all `.teaser__picture` instances, not just those appearing in the grid listing. Is this intended, and if so has it been cross browser tested in all teaser types?

A lighter touch would be to rather modify the selector so that it only doesn't apply to those images where a problem has been identified, e.g. something like `.teaser:not(:teaser--grid-style) .teaser__picture img` (untested).

discodavey:
I have tested all the classes with teasers as there is a page in molecules/teasers/view all. Both your way and what I did works exactly in the same way. There is a slight change in the square images on removing the code as it moves the image over to the right by 1px.

davidcmoulton:
Cool. This is the image resizing from `auto` to `100%` and is the desired effect.

PR #935, opened 2019-01-17
Remove focus

discodavey:
Looks to be causing anchor placing issues

### DIFF
--- a/assets/js/components/Popup.js
+++ b/assets/js/components/Popup.js
@@ -239,6 +239,8 @@ module.exports = class Popup {
       this.popupHitBox.setAttribute('aria-expanded', 'true');
       this.popupHitBox.setAttribute('tabindex', -1);
       this.popupHitBox.setAttribute('aria-hidden', 'false');
+      this.popupHitBox.focus();
+      this.preventDefault();
     } else {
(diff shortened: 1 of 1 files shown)


discodavey:
I have removed  ```this.popupHitBox.focus();```  for the time being so this will correct the issue for the moment. We should look at how we implement the focus later on with more testing to make sure it does not have any implications on anything else.

davidcmoulton:
How does removing this focus affect the accessibility of the experience?

discodavey:
It allows the anchor to work correctly and move down the page where as if this is in place it does not trigger the anchor. This is what happens on the tests I have done on my mac when in localhost.

davidcmoulton:
Does this change alter the screen reader / keyboard navigation experience?

discodavey:
Yes, As in my comment above I think we should remove this line of code for the time being so the pop up's work correctly and then look into the focus element and see what we can do to implement it without causing issues.

It will alter the experience but will not stop them from navigating the page. If I remember correctly without this line of JS it does not focus all the time but in general it was better than what they had before I put all the code in.

discodavey:
Tested on NVDA and works correctly.

PR #936, opened 2019-01-17
Amend focus colour

discodavey:
Been tested for accessibility and passes


source/_patterns/01-molecules/components/article-section--kitchen-sink.mustache
```diff
@@ -82,8 +82,8 @@
         <tr>
           <td>Division 1</td>
           <td>Division 2</td>
-          <td>Division 3</td>
-          <td>Division 4</td>
+          <td><a href="">Division 3 with link</a></td>
```
davidcmoulton: `href="#"` is useful in this situation so if a user clicks on what should be a non-functional link, the browser doesn't load anything.

PR #939, opened 2019-01-21
Left align content in author popup

discodavey:


discodavey:
I have spoken to Nick and Chris H about the layout of whether there should be bullets or not. 

The decision has been made to left align the text for the time being and create a new ticket for the bullet aspect as changing removing the bullets and adjusting the spacing will also change the article and author information section.

This will need approval from Melissa and Peter if anything changes in this section.

New ticket for bullets issue is : https://github.com/elifesciences/issues/issues/4709

PR #940, opened 2019-02-01
js fix on pop up

discodavey:


discodavey:
As discussed yesterday (Thursday 31st) with preventDefault()

PR #946, opened 2019-03-04
Call to action pattern

davidcmoulton:
Still needs JavaScript.

### DIFF
--- a/assets/sass/patterns/molecules/call-to-action.scss
+++ b/assets/sass/patterns/molecules/call-to-action.scss
@@ -24,7 +24,16 @@
   grid-template-columns: 80px 1fr 12px auto;
   grid-template-rows: 1fr;
   grid-column-gap: 16px;
-  grid-row-gap: #{get-rem-from-px(24)}rem;
+  grid-row-gap: #{get-rem-from-px($blg-space-small-in-px)}rem;
+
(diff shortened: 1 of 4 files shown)


source/_patterns/01-molecules/components/call-to-action.mustache
```diff
@@ -0,0 +1,36 @@
+<div class="call-to-action-wrapper" id="{{id}}">
+
+  <div class="call-to-action">
+
+    {{#image}}
```
thewilkybarkid: Shouldn't be needed?

assets/sass/patterns/molecules/call-to-action.scss
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    grid-column: 3;
+    grid-row: 1;
+  }
+
+}
+
+.call-to-action__close {
+  // TODO: Update when js injects .call-to-action__close into the DOM instead of it being created by the template
```
thewilkybarkid: Can we just remove for now?

davidcmoulton: Gone.

source/_patterns/01-molecules/components/call-to-action.mustache
davidcmoulton: It's gone.

PR #954, opened 2019-09-25
Amend Curated to Edited

discodavey:


discodavey:
This is two parts of changing Curated to Edited. The other changes are in Journal.

PR #955, opened 2019-10-08
page views on charts/metrics bug

discodavey:


discodavey:
Chris H recommended this approach and has been approved by Chris and Maël.

assets/sass/patterns/atoms/triggers.scss
```diff
@@ -1,7 +1,6 @@
 @import "../../definitions";
 
 .trigger {
-  background-color: white;
```
davidcmoulton: Seems to work.

My only concern is that this makes the visibility of the arrow reliant on the colour of whatever it sits on top of.

discodavey: Having spoken with Chris H we had decided that this was an edge case as the chances of this happening on a regular basis would be minimal hence the reason to go down this route. With having the previous and next buttons at the bottom they seem to get lost.

PR #957, opened 2019-11-20
Amend so div does not fit 100% width of content header

discodavey:


discodavey:
This is now not filling 100% of the content header now so collections link can now be clicked

PR #958, opened 2019-12-02
Dismissible info-bar

discodavey:


discodavey:
@thewilkybarkid  That's correct. I have only done the pattern. Mael mentioned about you doing the remainder of the work as it was assumed it would be back end. Are we only just doing JS and cookies?

thewilkybarkid:
No, it's pretty much all JavaScript. Journal will add the banner to the right pages, it's up to JS to decide whether to show it or not.

assets/js/actions/Dismiss.js
```diff
@@ -3,9 +3,10 @@
 const utils = require('../libs/elife-utils')();
 
 module.exports = class Dismiss {
-  constructor($parent, $patternRoot, document, cookieExpiry) {
+  constructor($parent, $toDismiss, document, cookieExpiry) {
+    this.$toDismiss = $toDismiss;
     this.doc = document;
-    this.cookieName = Dismiss.deriveCookieName($patternRoot);
+    this.cookieName = Dismiss.deriveCookieName(this.$toDismiss);
```
discodavey: To dismiss what? I do not think this clear as what we are dismissing.

davidcmoulton: `$toDismiss` is the element to dismiss. We know it's an element because it's `$`-prefixed, according to the convention for this codebase. It's a more meaningful name than `$patternRoot`, as it tells you why it's there, rather than just what it is.

source/_patterns/00-atoms/components/info-bar.yaml
```diff
@@ -8,11 +8,15 @@ properties:
     type: string
     enum:
       - attention
+      - dismissible
       - info
       - multiple-versions
       - correction
       - success
       - warning
+  dismissibleOnOff:
```
giorgiosironi: using `OnOff` as a suffix is a bit confusing

source/_patterns/00-atoms/components/info-bar.yaml
```diff
@@ -8,11 +8,15 @@ properties:
...
+    type: boolean
+    default: false
```
giorgiosironi: several properties like `cookieDuration` appear in `info-bar~*.json` examples but not in this schema?

assets/sass/patterns/atoms/dismiss-button.scss
```diff
@@ -0,0 +1,44 @@
+@import "../../definitions";
+
+.dismiss-button {
+
+  .call-to-action & {
```
davidcmoulton: `dismiss-button` is not a pattern so should not have its own sass file. The separate implementations in this file, `.call-to-action` and `.info-bar`, are very different and there is no need to group them together.

assets/js/components/CallToAction.js
```diff
@@ -1,77 +1,20 @@
...
-    this.cookieName = CallToAction.deriveCookieName(this.$elm);
-    if (!this.cookieName.length || this.hasAlreadyBeenSeen(this.cookieName)) {
-      this.hide();
-      return;
-    }
-
-    this.$button = CallToAction.buildDismissButton();
-    this.$elm.addEventListener('click', this.handleInteraction.bind(this));
+    this.dismissible = new Dismissible(this.$elm, '.call-to-action', this.doc);
+    this.$elm.addEventListener('click', () => this.dismissible.dismiss());
```
giorgiosironi: doesn't this duplicate https://github.com/elifesciences/pattern-library/pull/958/files#diff-b24b40c83ce03b33cefdfd2daffdc65eR56 where the button has its own event listener?
Also, does it mean that clicking anywhere on the CallToAction triggers the dismissal?

source/_patterns/00-atoms/components/info-bar.mustache
```diff
@@ -1,3 +1,14 @@
-<div class="info-bar info-bar--{{type}}">
-  <span class="info-bar__text">{{{text}}}</span>
+<div class="info-bar info-bar--{{type}}" id="{{id}}"
+     {{#dismiss}}
```
giorgiosironi: I see this will reflect in `patterns-php` providing one more attribute to configure rather than a different `type` for this InfoBar

source/_patterns/00-atoms/components/info-bar.mustache
```diff
@@ -1,3 +1,14 @@
...
+        data-behaviour="InfoBar"
+        {{#cookieDuration}} data-cookie-name-root="infoBar_" data-cookie-duration="{{cookieDuration}}"{{/cookieDuration}}
+        {{#cookieExpires}} data-behaviour="InfoBar" data-cookie-name-root="infoBar_" data-cookie-expires="{{cookieExpires}}"{{/cookieExpires}}>
```
giorgiosironi: `data-behaviour="InfoBar"` duplicated?

source/_patterns/00-atoms/components/info-bar.yaml
```diff
@@ -8,11 +8,28 @@ properties:
     type: string
     enum:
       - attention
+      - dismissible
```
giorgiosironi: this is not necessary anymore with the separate `dismiss` property?

giorgiosironi: Assuming absence of the property means not dismissible, and presence of the property means dismissible; this makes one of the `cookie*` properties mandatory, which is in fact captured in this schema

test/calltoaction.spec.js
```diff
@@ -1,151 +1,133 @@
...
-          $elm.parentElement.appendChild($elmNoId);
+          it('is configured with a day offset of the value of data-cookie-duration HTML attribute, if set', function () {
+            const id = fixtures.generateRandomId();
 
-          callToActionWithElmNoId = new CallToAction($elmNoId);
-        });
+            // 7 days in the future
+            let expiry = new Date(new Date());
+            expiry.setDate(expiry.getDate() + 7);
+            expiry = expiry.toUTCString();
```
giorgiosironi: given the second precision, this may be vulnerable to race conditions where the tests expected values becomes slightly off if the test code and the production code run too far apart. Any way to ~mock~ stub `Date`?

source/_patterns/00-atoms/components/info-bar.mustache
davidcmoulton: The different type is still needed as a differentiator in the css selectors.

source/_patterns/00-atoms/components/info-bar.yaml
davidcmoulton: This value is [used in the css selectors](https://github.com/elifesciences/pattern-library/pull/958/files#diff-007b18257254acbb20bed8144ad26566R78).

source/_patterns/00-atoms/components/info-bar.mustache
giorgiosironi: Any way to constrain the schema to say that if `type` is `dismissible` than `dismiss` is required?

assets/js/components/CallToAction.js
davidcmoulton: This should be attached to the button of the call to action pattern, rather than the entire thing, to dismiss it once it's actioned. You're right that currently it is dismissed too keenly, when any part of the pattern is clicked on.

source/_patterns/00-atoms/components/info-bar.mustache
davidcmoulton: I've tried to do this before, but I don't believe that it's possible to set a required property based on the value of another property.

source/_patterns/00-atoms/components/info-bar.mustache
davidcmoulton: Removed in https://github.com/elifesciences/pattern-library/pull/958/commits/8a19524ed9f55735cb986edac6152ddb27800b21

assets/js/components/CallToAction.js
davidcmoulton: Updated in https://github.com/elifesciences/pattern-library/pull/958/commits/da42b336745f9fd75147d9a705f4f13998043c8e

source/_patterns/01-molecules/components/call-to-action.mustache
```diff
@@ -1,4 +1,15 @@
-<div class="call-to-action-wrapper{{#needsJs}} call-to-action-wrapper--needs-js{{/needsJs}}" id="{{id}}" data-behaviour="CallToAction">
+{{! Either cookieDuration, or cookie expiry date, not both }}
```
giorgiosironi: outdated comment?

test/calltoaction.spec.js
davidcmoulton: Removed the underlying feature of setting `cookieDuration`, and so also this test.

assets/js/components/_Dismissible.js
```diff
@@ -0,0 +1,69 @@
...
+    if (this.hasCookieName(this.cookieName) && this.hasPreviouslyBeenDismissed(this.cookieName, this.doc.cookie)) {
+      this.hide();
+      return;
+    }
+
+    this.cookieExpiryDate = Dismissible.setCookieExpiryDate(this.$toDismiss);
+    this.$button = this.buildButton(attachPoint);
+  }
+
+  static setCookieExpiryDate($elm) {
```
giorgiosironi: a `set*` that returns rather than modifying `this` is a bit unexpected, but maybe I'm not aware of a naming convention in use here

source/_patterns/01-molecules/components/call-to-action.yaml
```diff
@@ -20,3 +27,7 @@ required:
   - text
   - button
   - image
+dependencies:
+  dismiss:
```
giorgiosironi: shouldn't this be `dismissible` matching the field name?

test/calltoaction.spec.js
```diff
@@ -1,151 +1,114 @@
...
-      });
+          it('is derived from the HTML element\'s id and data-cookie-name-root attribute', function () {
+            const id = fixtures.generateRandomId();
+            const $fixture = fixtures.generateHTMLWithCookieDetails(fixtures.getCookieNameRoot(), id);
+            const callToAction = new CallToAction($fixture);
+            callToAction.dismissible.$button.click();
+            const expectedCookieName = `${fixtures.getCookieNameRoot()}${id}`;
+            expect(callToAction.dismissible.cookieName).to.equal(expectedCookieName);
+            expect(utils.getCookieValue(expectedCookieName, document.cookie)).to.equal('true');
+            fixtures.clearCookie(id);
```
giorgiosironi: not sure if a failed `expect` interrupts the test, but shoudn't clean up code be in an `afterEach()` in general? (wrapped in an `if` in case it were not always needed)

test/calltoaction.spec.js
```diff
@@ -1,151 +1,114 @@
...
-          $elmNoId.parentElement.removeChild($elmNoId);
-        });
+        describe('expiry date', function () {
 
-        it('does not create a dismiss button', () => {
-          expect(callToActionWithElmNoId.$button).to.be.undefined;
-        });
+          it('is configured with the value of data-cookie-expires HTML attribute, if set', function () {
+            const id = fixtures.generateRandomId();
+            const expiry = 'Tue, 19 January 2038 03:14:07 UTC';
```
giorgiosironi: this is just a string passed through so the code doesn't really care about the date format, but isn't a the format `Tue, 19 Jan 2038 03:14:07 UTC`? (looking at the `Expires` example in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie)

test/info-bar.spec.js
```diff
@@ -0,0 +1,119 @@
...
+            $fixture.parentElement.removeChild($fixture);
+          });
+
+        });
+
+        describe('expiry date', function () {
+
+          it('is configured with the value of data-cookie-expires HTML attribute, if set', function () {
+            const id = fixtures.generateRandomId();
+            const expiry = 'Tue, 19 January 2038 03:14:07 UTC';
```
giorgiosironi: (similar month format to check?)

assets/js/components/_Dismissible.js
giorgiosironi: (for now I'll just leave it)

test/calltoaction.spec.js
giorgiosironi: Not modifying this without understanding the reason for its placement.

giorgiosironi: Looking at the other tests, it looks more like a missing `expect`.

PR #959, opened 2020-01-17
Change line height to allow for url underline

discodavey:


assets/sass/patterns/atoms/info-bars.scss
```diff
@@ -85,6 +85,8 @@
 
   .info-bar__text {
 
+    @include font-size-and-vertical-height(14, 22.4);
```
giorgiosironi: corresponds to a tested `1.6` blessed by the UX designer

PR #962, opened 2020-01-27
changed up arrow to center from left aligned

discodavey:


discodavey:
As well as centering the arrow we have added 12px to the top to make it more obvious away from the content. This was suggested and agreed with Chris W and Mael.

thewilkybarkid:
Chris H? 😅

discodavey:
Sorry. Have now corrected.

assets/sass/patterns/organisms/content-header.scss
```diff
@@ -501,7 +501,7 @@ $iconlist: cc oa;
 }
 
 .content-header__author_link_highlight {
-  padding: 0;
+  padding: 12px 0 0;
```
davidcmoulton: Need to remove this padding at wider viewport sizes to prevent this extra vertical space:

Screenshot 2020-01-28 09 30 24

davidcmoulton: Do the right / left / bottom values need to be set to 0? If not, just use `padding-top`.

discodavey: The reason I have coded it in the way is that padding stands out more than just padding-top so if any other developer reads it and needs to add more padding to the left/right or bottom then it's just a case of adding it. 

I find it easier to read.

davidcmoulton: If the code is correct then it makes sense to try to improve the developer experience. Including rules that aren't needed isn't correct though, and can cause problems down the line. Please use `padding-top` here.

discodavey: Just my opinion

assets/sass/patterns/organisms/content-header.scss
```diff
@@ -500,15 +500,16 @@ $iconlist: cc oa;
...
-}
-
 .content-header__author_link_highlight,
 .content-header__author_link_highlight:hover {
   background-color: transparent;
   border-style: none;
   color: $color-primary;
+  padding-top: 0;
+
+  @media only all and (max-width: #{get-rem-from-px($bkpt-site--medium)}em) {
```
davidcmoulton: Media queries should be `min-width`, so set the `padding-top` for narrow screens in `.content-header__author_link_highlight` and zero it in the media query.

PR #964, opened 2020-02-10
Allow arbitrary links in the ViewSelector

thewilkybarkid:
Avoiding a full-on refactor, this adds an `otherLinks` list to the ViewSelector.

https://github.com/elifesciences/issues/issues/5402

### DIFF
--- a/assets/js/components/ViewSelector.js
+++ b/assets/js/components/ViewSelector.js
@@ -235,7 +235,12 @@ module.exports = class ViewSelector {
 
   insertSideBySideListItem() {
     const $list = this.doc.querySelector('.view-selector__list');
-    const $jumpList = this.doc.querySelector('.view-selector__list-item--jump');
+    const $jumpList = this.doc.querySelector('.view-selector__list-item--figures + .view-selector__list-item');
+
+    if (!$jumpList) {
+      return;
+    }
+
     const $listItem = utils.buildElement(
       'li',
       [
--- a/assets/js/components/ViewSelector.js
+++ b/assets/js/components/ViewSelector.js
@@ -235,9 +235,9 @@ module.exports = class ViewSelector {
 
   insertSideBySideListItem() {
     const $list = this.doc.querySelector('.view-selector__list');
-    const $jumpList = this.doc.querySelector('.view-selector__list-item--figures + .view-selector__list-item');
+    const $attachBefore = this.doc.querySelector('.view-selector__list-item--figures + .view-selector__list-item');
 
-    if (!$jumpList) {
+    if (!$attachBefore) {
       return;
     }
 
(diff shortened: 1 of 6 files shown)


assets/js/components/ViewSelector.js
```diff
@@ -235,7 +235,12 @@ module.exports = class ViewSelector {
 
   insertSideBySideListItem() {
     const $list = this.doc.querySelector('.view-selector__list');
-    const $jumpList = this.doc.querySelector('.view-selector__list-item--jump');
+    const $jumpList = this.doc.querySelector('.view-selector__list-item--figures + .view-selector__list-item');
```
giorgiosironi: it's not the jump list anymore but may be the repro link? Is it just the `$attachBefore` argument now?

source/_patterns/01-molecules/navigation/view-selector.json
```diff
@@ -2,6 +2,9 @@
   "articleUrl": "#article",
   "figureUrl": "#figures",
   "sideBySideUrl": "#sideBySide",
+  "otherLinks": [
+    { "name": "Executable code", "url": "#rds" }
```
giorgiosironi: schema matches `jumpLinks.links`

assets/sass/patterns/molecules/view-selector.scss
```diff
@@ -125,6 +125,7 @@
   }
 
   .view-selector__list-item {
+    display: none;
```
giorgiosironi: what makes the new item visible then?

thewilkybarkid: https://github.com/elifesciences/pattern-library/blob/150c6f2af10fba2278e0addef7a21b8519f5bd3f/assets/sass/patterns/molecules/view-selector.scss#L138-L139 and https://github.com/elifesciences/pattern-library/blob/150c6f2af10fba2278e0addef7a21b8519f5bd3f/assets/sass/patterns/molecules/view-selector.scss#L144-L145

giorgiosironi: Got it for the first 2 items, but I am missing how the `<li class="view-selector__list-item">` from the `otherLinks` part of the template get the same treatment, unless one of these classes is added through JS (but they seem too specific)?

thewilkybarkid: Pre-existing in the template: https://github.com/elifesciences/pattern-library/blob/150c6f2af10fba2278e0addef7a21b8519f5bd3f/source/_patterns/01-molecules/navigation/view-selector.mustache#L3-L10

(This is the part I've fought refactoring.)

giorgiosironi: Figured it out: this is `none` because it's the selector for smaller screens and there the link disappears :man_facepalming:

PR #965, opened 2020-02-13
Remove hard-coding of teaser formats

thewilkybarkid:
Allows the number/wording to be changed.

https://github.com/elifesciences/issues/issues/5416

(I could have sworn it already worked like this, maybe we just talked about it and didn't do it.)

### DIFF
--- a/source/_patterns/01-molecules/teasers/teaser.yaml
+++ b/source/_patterns/01-molecules/teasers/teaser.yaml
@@ -85,12 +85,14 @@ properties:
         $ref: ../../01-molecules/components/meta.yaml
       formats:
         type: object
-        minProperties: 1
         properties:
-          html:
-            type: boolean
-          pdf:
-            type: boolean
+          list:
+            type: array
+            minItems: 1
+            item:
+              type: string
+        required:
+          - list
     required:
       - meta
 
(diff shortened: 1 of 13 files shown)


source/_patterns/04-pages/home.json
```diff
@@ -377,8 +377,7 @@
               }
             },
             "formats": {
-              "html": true,
-              "pdf": true
+              "list": ["Executable", "PDF"]
```
thewilkybarkid: 👋

source/_patterns/01-molecules/teasers/teaser.yaml
```diff
@@ -85,12 +85,14 @@ properties:
...
         properties:
-          html:
-            type: boolean
-          pdf:
-            type: boolean
+          list:
+            type: array
+            minItems: 1
+            item:
+              type: string
```
giorgiosironi: not super future-proof if something like a description gets added (or any metadata). May be good for now given how relatively easy this schema can change.

source/_patterns/04-pages/home.json
giorgiosironi: shouldn't this also have `HTML`?

giorgiosironi: I know this is just an example, but is the uppercase implemented via CSS?

giorgiosironi: > shouldn't this also have HTML?

Ah re-read the whole issue, there's... not enough space for that? Whatever, the example is intended.

PR #967, opened 2020-02-14
Introduce template for embedded tweets

nlisgo:


source/_patterns/00-atoms/components/tweet.yaml
```diff
@@ -0,0 +1,31 @@
...
+    type: string
+    minLength: 1
+  text:
+    type: array
+    items:
+      $ref: ../../00-atoms/components/paragraph.yaml
+    minItems: 1
+  date:
+    $ref: ../../00-atoms/components/date.yaml
+  hideConversation:
```
davidcmoulton: For consistency, prefer to use the `enum` approach for defining booleans, e.g. 

```YAML
hideConversation:
  type: boolean
  enum:
    - true
```
Also [see further example](https://github.com/elifesciences/pattern-library/blob/develop/source/_patterns/00-atoms/components/speech-bubble.yaml#L10 for example).

This has the advantage of if it's present, it's true, which plays nicely with mustache.

source/_patterns/00-atoms/components/tweet.mustache
```diff
@@ -0,0 +1,13 @@
...
+    {{#text}}
+      {{> atoms-paragraph }}
+    {{/text}}
+    &mdash; {{accountLabel}} (@{{accountId}}) <a href="{{url}}">
+    {{#date}}
+      {{>atoms-date}}
+    {{/date}}
+    </a>
+  </blockquote>
+  <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
```
nlisgo: Wasn't sure which was the favoured approach to introduce this library, so it was introduced like this to get things working quickly.

davidcmoulton: 👍 For the actual implementation, it might be worth looking at [MathJax is handled].(https://github.com/elifesciences/pattern-library/blob/develop/assets/js/components/Math.js)

PR #968, opened 2020-02-21
Amend class name

discodavey:
Third party changed class name to hypothesis-highlight from annotator-hl


thewilkybarkid:
https://github.com/hypothesis/client/pull/1775 is the related change.

PR #974, opened 2020-03-11
Introduce template for embedded google map

nlisgo:
Improves: https://github.com/elifesciences/issues/issues/5497

The following content we have added an image of map and a link to the interactive map: https://elifesciences.org/inside-elife/a946c355/elife-community-ambassadors-243-volunteers-join-the-programme-in-2019

The solution implemented embeds this map into the page as an iframe.


nlisgo:
Screenshot 2020-03-11 at 16 48 26

PR #982, opened 2020-05-05
Fix so links can be clicked in popup

discodavey:


discodavey:
This is the test link for the pattern library.

https://s3.amazonaws.com/ci-pattern-library/982/patterns/04-pages-article--research/04-pages-article--research.html

davidcmoulton:
Clicking an author name doesn't show a popup.

PR #985, opened 2020-05-14
Add example article with structured abstract

nlisgo:


nlisgo:
Once this is merged into master I will publish the npm package manually.

PR #989, opened 2020-06-17
Amend text to be consistent/correct

discodavey:


source/_patterns/04-pages/article--research~with-structured-abstract.json
```diff
@@ -590,11 +590,11 @@
             },
             {
               "primary": {
-                "name": "Executable DAR",
+                "name": "Executable version",
                 "url": "#darLink"
```
nlisgo, line 594: nit pick: let's change this to #eraLink as it will no longer be DAR files.

nlisgo: Optionally, you could also just remove the links to the `Executable version` from this example article so that we just have one example? No preference.

PR #994, opened 2020-07-31
Remove papers from downloads links

discodavey:


test/articledownloadlinkslist.html
```diff
@@ -84,10 +84,7 @@ <h3 class="article-download-links-list__heading">Open citations<span class="visu
 
 
         >Papers App</a></li>
-        <li><a href="#" class="article-download-links-list__link"
-
-
-        >CiteUlike</a></li>
+
```
nlisgo, line 87: Just need to remove `Papers App` link from this one also

NuclearRedeye:
@discodavey What's the status of this PR?

discodavey:
Looks like it has been removed from production already. Just needs to be done from the pattern library.

This is what it looks like on production
https://elifesciences.org/articles/61920

but on the pattern library it is still there so this needs to be merged.
https://ui-patterns.elifesciences.org/patterns/04-pages-article--research/04-pages-article--research.html

Add it to the board and I can do it today.

PR #1007, opened 2020-09-04
Jump menu expanded

discodavey:


test/viewselector.spec.js
```diff
@@ -329,23 +305,6 @@ describe('A ViewSelector Component', function () {
 
     });
 
-    it('maintains a list of top level article section headings', () => {
```
NuclearRedeye, line 332: We might want to raise a follow up ticket just to ensure that this test is covered somewhere else. If the other tests are only checking if the menu has the default options and not the section headings links then it is possible we might miss a break further down the line.

NuclearRedeye (approved):
I've made one comment, that I think we should investigate as part of some follow up work. I'm still annoyed that I cannot get your changes to run in a local instance (not your fault!), but I've tested them on the AWS instance and am happy that they are working as expected and no regressions have been introduced.

discodavey:
@NuclearRedeye So you are happy for me to put this live then?

NuclearRedeye:
@discodavey Yep

PR #1008, opened 2020-09-09
Change css class for button

discodavey:


discodavey:
This is in reference to issue #6045

nlisgo (approved):

gnott:
If PR https://github.com/elifesciences/patterns-php/pull/745 is acceptable, then we can keep the `button--variation` class name, and we wouldn't need to change it.

PR #1011, opened 2020-09-15
Allow image of personalised cover template to be overidden

nlisgo:


source/_patterns/01-molecules/components/personalised-cover-download.mustache
```diff
@@ -25,12 +25,21 @@
...
     </div>
     <div class="personalised-cover-thumbnail">
-      <picture>
-        <source srcset="{{#assetRewrite}}{{assetsPath}}/img/patterns/molecules/cover-preview.jpg{{/assetRewrite}}" type="image/svg+xml">
-        <img srcset="{{#assetRewrite}}{{assetsPath}}/img/patterns/molecules/cover-preview.jpg{{/assetRewrite}} 276w"
-          src="{{#assetRewrite}}{{assetsPath}}/img/patterns/molecules/cover-preview.jpg{{/assetRewrite}}" sizes="138px"
-          alt="" class="personalised-cover-download__icon">
-      </picture>
+      {{#image}}
+        <picture class="profile-snippet__picture">
```
gnott, line 29: This class name could possibly be more specific to personalised covers, or removed?

gnott: Testing it too, it isn't suitable because then it has `max-width: 4.5rem;` and the image appears too small, so remove or new class name? Thanks!

gnott:
Testing locally, I think this template will work. There's a weird border at the bottom of the image, but this might be due to other factors:

image

It looks like tests are failing due to the ` 276w` in the fallback JSON data is refused by the schema, but the personalised covers project does not use the fallback and is unaffected.

Thanks for this @nlisgo!

gnott:
Correcting my comment above, it looks like fallback does get used, but it's all good.

Another thing after testing more, if we add ` class="personalised-cover-download__icon"` to the `` tag then we can retain more styles. I will update the file in this branch directly with that, I hope it's ok to collaborate on this branch with you.

gnott:
Here is what I am able to produce so far based on the current strategy (regarding the `1x` image resolution):

```
<div class="personalised-cover-thumbnail">
        <picture>
            <source srcset="http://127.0.0.1:5000/elifecovers/15278-cover-a4-thumb.jpg">
            <img src="http://127.0.0.1:5000/elifecovers/15278-cover-a4-thumb.jpg" srcset="http://127.0.0.1:5000/elifecovers/15278-cover-a4-thumb.jpg 1x" alt="" class="personalised-cover-download__icon">
        </picture>
    </div>
```

discodavey (approved):
Looks all ok. Are you going to set up a test for this?

nlisgo:
> Looks all ok. Are you going to set up a test for this?

A test for the javascript specifically? If you are happy with the approach then I can look at adding a test if appropriate. Will look at other examples.

discodavey:
Would be good if we could do a test for javascript if appropriate.

assets/sass/patterns/molecules/personalised-cover-download.scss
```diff
@@ -54,27 +66,26 @@
...
   display: grid;
-  grid-row: 2;
   justify-items: center;
 }
 
-.button-collection__item {
-  @include margin(12, "left");
-  @include margin(12, "right");
-}
+ol.personalised-cover-buttons[data-behaviour-initialised="true"] {
```
nlisgo, line 72: I believe this should be `.personalised-cover-buttons[data-behaviour-initialised="true"]`

assets/sass/patterns/molecules/personalised-cover-download.scss
```diff
@@ -54,27 +66,26 @@
...
+    @include margin(12, "right");
+  }
 }
 
-.personalised__button {
-  @include colors($color-primary);
-  @include padding(18 18 14);
-  font-weight: 500;
-  letter-spacing: 0.4px;
+ol.personalised-cover-buttons[data-behaviour-initialised="false"] {
```
nlisgo, line 84: I believe this should be `.personalised-cover-buttons[data-behaviour-initialised="false"]`

discodavey: I agree and have moved them into .personalised-cover-buttons class to reduce the code. Works for me so please can you test?

nlisgo:
I may introduce the selenium test as a followup.

nlisgo:
@discodavey test added.

test/personalisedcoverdownload.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    let a4ButtonCollection;
+    let letterButtonCollection;
+
+    before(() => {
+      a4ButtonCollection = $elm.querySelector('.button-collection--a4');
+      letterButtonCollection = $elm.querySelector('.button-collection--usletter');
+    });
+
+    context('default buttons', () => {
+      it('the A4 download menu is not hidden', () => {
```
discodavey, line 30: In previous tests we have made these so they read like sentences so for example:

"When A4 has been selected"

"the A4 download menu is not hidden"

"and the US letter download menu is hidden"

Do you see what I mean or are we changing the way we are doing our tests?

nlisgo: Have reworded the descriptions

PR #1017, opened 2020-09-17
Add capitialize meta type

discodavey:


discodavey:
Just reverting this as was a pattern library issue and not journal

PR #1020, opened 2020-10-09
Hypothesis redesign

discodavey:


discodavey:
Closing as revising how this is built after speaking to @nlisgo

PR #1021, opened 2020-10-09
Make personalised cover download title changeable

nlisgo:


source/_patterns/01-molecules/components/personalised-cover-download.yaml
```diff
@@ -4,15 +4,6 @@ properties:
   title:
     type: string
     minLength: 1
-  headingLevel:
```
nlisgo, line 7: These 3 parameters weren't being referenced in the template so I removed them

nlisgo:
@discodavey I want to merge this. I will address any issues in a followup next week.

PR #1023, opened 2020-10-16
Hypothesis button move in article

discodavey:


discodavey:
This has been done with articles in mind. There is another ticket which takes other sections like digest etc in mind where they are set in different positions. Not sure if this is the correct approach as we need to take everything into consideration and think to be honest we should possibly make this just one ticket.

This is purely down to if we change article now if will have an effect on the rest of them as they are sat in a different positon on the page so will need to be done all at the same time.

@NuclearRedeye What do you think?

NuclearRedeye (approved):
Not tested locally, but changes seem OK to me

discodavey:
Although the ticket for this has been closed this is not to be merged until the rest of the work for the epic has been done.

PR #1025, opened 2020-10-22
Move Hypothesis controls in articles

NuclearRedeye:


discodavey:
There are only three pages that can be tested which are:
Position by first section - Default
Position centrally inline - inside eLife
Position end - digest

There is one more which position at the end of the second section which does not have page that is relevant to the code so this will need to be added to the list of things needed to be added to the pattern library.

@NuclearRedeye Can you please check you are happy with this and then we can revert the Hypothesis js file.

discodavey:
I cannot seem to select you for reviewing @NuclearRedeye

source/_patterns/04-pages/article--research.mustache
```diff
@@ -1,4 +1,4 @@
-<div class="global-wrapper">
+<div class="global-wrapper"  data-item-type="default">
```
NuclearRedeye, line 1: I know they result in the same output, but to match whats on prod it should be 'research-article'.

NuclearRedeye (commented):
So some further changes are needed...
1. I thought on line 25 of `article--research.mustache` the class `content-container` needs to be removed right? Or is that no longer the case?
2. What about adding the `data-item-type` to the other templates?

NuclearRedeye:
@discodavey That's because I'm the author of the PR. Can't be assigned as a reviewer to my own code! See my comments above, I think a few little changes are needed.

discodavey:
@NuclearRedeye The other data types are in the js but we only have three pages set up that can take the annotations data type. I do not think we need to put it in every version of article research. There are other pages which are not in the pages section of the pattern library and these can be done as another ticket as I do not feel this is something that should be done now and can be done as part of pattern library update.

source/_patterns/04-pages/article--research.mustache
discodavey: I have changed the one for the article section.

PR #1026, opened 2020-10-26
Add styling for attribution credit

discodavey:


discodavey:
Waiting for sign off from Chris H so I can then add the json to the component.

discodavey:
This is for the pattern library not journal side.

source/_patterns/01-molecules/components/carousel-item.yaml
```diff
@@ -24,6 +24,9 @@ properties:
   title:
```
NuclearRedeye, line 24: I can't view this molecule standalone, just have a blank page. Should I be able to?

Edit: Actually I can't seem to get any of the Carousel molecules to display be selecting them from the `molecules->components` menu. Other molecules work, so I assume this should too?

NuclearRedeye (commented):
Test locally, when I can load the pages it works well. But some items aren't working. Not sure if this is a regression, if so it should be fixed. If not, then we should raise another ticket to cover fixing those.

source/_patterns/01-molecules/components/carousel-item.yaml
NuclearRedeye: So I just tested this locally without your changes and it's still broken, so not a regression as a result of your change. I'll raise a separate issue to cover fixing them.

NuclearRedeye (approved):

PR #1028, opened 2020-11-02
adding classes for GTM to pick up

discodavey:


discodavey:
Added the classes so we can get Tag Manager to register the click correctly.

NuclearRedeye (approved):
Changes seem good to me

PR #1030, opened 2020-11-07
Allow variant of speech bubble which is wrapped

nlisgo:
I believe this is a better approach for introducing the design changes needed for the wrapped speech bubble in https://github.com/elifesciences/issues/issues/6095

This avoids injecting any text into the template and it could feasibly be used in other places. I'm not sure I'm happy with using the word `wrapped` another option would be `inverted` but we can agree that when we pair next week.

If the `isWrapped` flag is not set, then the prefix text just gets added within the speech bubble.


nlisgo:
All other variants of `speech-bubble` should appear as before and the new variant should appear as expected: https://s3.amazonaws.com/ci-pattern-library/1030/patterns/00-atoms-components-speech-bubble-share-your-feedback/00-atoms-components-speech-bubble-share-your-feedback.html

PR #1032, opened 2020-11-10
Speech bubble redesign extended

nlisgo:


nlisgo:
All that remains is to extend the test coverage to cover the 3 new functions I've added to `HypothesisOpener`:

- `targetWithinInLineProfile`
- `targetWithinBlockquote`
- `targetBelowSectionHeading`

assets/js/components/HypothesisOpener.js
```diff
@@ -203,9 +203,25 @@ module.exports = class HypothesisOpener {
...
+        $elm.style.marginRight = 'auto';
+      } else {
+        $elm.style.float = 'right';
+        $elm.style.marginLeft = 'unset';
+        $elm.style.marginRight = 'unset';
+      }
+
+    }
+
+    window.addEventListener('resize', annotationPosition);
```
nlisgo, line 224: why does any of this need to be in javascript. could we not just add the css?

NuclearRedeye: It would probably be better to just toggle classes, but let's not worry about resolving that here.

assets/js/components/HypothesisOpener.js
```diff
@@ -330,6 +351,39 @@ module.exports = class HypothesisOpener {
     return positioners[articleType] || positioners.default;
   }
 
+  static findTarget($contentContainer) {
+    let target = null;
+    const centre = Math.floor(($contentContainer.querySelectorAll('p').length - 1) / 2);
+    [].forEach.call($contentContainer.querySelectorAll('p'), (possible, index) => {
```
NuclearRedeye, line 357: Not that i'd worry about this now, but this could be not improved so to say but neatened up...

```
target = $contentContainer.querySelectorAll('p').find( (possible) => {
  if (
        target === null &&
        index >= centre &&
        !HypothesisOpener.targetWithinInLineProfile(possible) &&
        !HypothesisOpener.targetWithinBlockquote(possible) &&
        !HypothesisOpener.targetBelowSectionHeading(possible)
      ) {
        return true;
      }
  return false;
});
```

NuclearRedeye (approved):
Some potential for minor optimisation, but nothing major that this should be held up.

assets/js/components/HypothesisOpener.js
nlisgo: I tried this but it doesn't seem to work on the output of `$contentContainer.querySelectorAll('p').find`. I think I would need to convert `$contentContainer.querySelectorAll('p')` to an array and this would risk losing the context of the elements. Since we need to interrogate the parentNodes of these elements we can't afford to lose the context. I wish this would work as it is much cleaner than my code.

NuclearRedeye: Yep, my bad. Seems it returns a [nodelist](https://developer.mozilla.org/en-US/docs/Web/API/NodeList) rather than an array and hence the filter isn't valid. Standard loop would work though...
```
for (const possible in $contentContainer.querySelectorAll('p')) {
    if ( index >= centre &&
        !HypothesisOpener.targetWithinInLineProfile(possible) &&
        !HypothesisOpener.targetWithinBlockquote(possible) &&
        !HypothesisOpener.targetBelowSectionHeading(possible)
      ) {
        target = possible;
        break;
      }
}
```

NuclearRedeye: Honestly though, just commit it as is.

assets/js/components/HypothesisOpener.js
nlisgo: > It would probably be better to just toggle classes, but let's not worry about resolving that here.

I managed to move this to css within a media query

PR #1033, opened 2020-11-14
remove speech bubble when failure triggered

nlisgo:


NuclearRedeye (commented):
I've checked this out and tried it locally. I had to make a modification locally to get the below path to trigger, which worked as expected. Note, I think the modification _might_ be worth committing, hence be good to discuss.

NuclearRedeye (approved):
Tested locally and work as expected. I'm happy for these changes to be merged

PR #1034, opened 2020-11-16
Sort outlines in carousel

discodavey:


discodavey:
I have removed the tabindexes for the time being. We will need to look into the accessibility side of things as to why this was done as you can tab through the carousel ok.

NuclearRedeye (approved):
Tested locally, this does resolve the issue with focus on the carousel but having looked at this JS in more detail I think I'm beginning to understand what the original intent of the code is. It's actually broken though, and not doing what it should be. Out of scope of this ticket however.
Do raise a followup ticket about the Aria tags though, and then based on the longer term goal for the carousel we can determine if it is worth investing the time to fix it or not.

PR #1043, opened 2020-12-17
Add word-break for long urls

discodavey:


NuclearRedeye (approved):
Tested locally, and looks good to me.

PR #1046, opened 2020-12-17
Alternative version of breaking long urls

discodavey:


assets/sass/patterns/molecules/article-section.scss
```diff
@@ -9,6 +9,13 @@ $arrow-size: 5;
   }
 }
 
+.article-section {
+
+  p > a {
```
NuclearRedeye, line 14: I'm no SASS expert, but should this be moved into the block above? e.g...
```
.article-section {
  &--collapsed {
    padding-bottom: 0;
  }
  p > a {
    overflow-wrap: break-word;
  }
}

```

NuclearRedeye (approved):

PR #1047, opened 2020-12-21
Hypothesis CTA at end of article for inside-elife, interview and press-package

nlisgo:


discodavey (approved):
All looks good from what I can test within the pages available to me.

PR #1048, opened 2020-12-21
Create new grid listing as breakpoint has changed on About profiles

discodavey:


discodavey:
We have had to create a new grid listing for the About profiles from the grid-listing class as the breakpoint is needed at 480 rather than 730px.

assets/sass/patterns/organisms/about-profiles.scss
```diff
@@ -1,14 +1,63 @@
...
 }
 
+.grid-listing__about {
+  @media only all and (min-width: #{get-rem-from-px($bkpt-site--small)}em) {
+    grid-template-columns: repeat(2, 1fr);
+    grid-column-gap: 2.25rem;
+    grid-row-gap: #{get-rem-from-px($blg-space-smallish-in-px)}rem;
+    justify-content: center;
+
+    // Supporting clients don't need clearfix
```
nlisgo, line 43: Is this meant to say, "Supporting clients that don't need clearfix"?

assets/js/components/AboutProfiles.js
```diff
@@ -11,11 +11,8 @@ module.exports = class AboutProfiles {
     this.window = _window;
     this.doc = doc;
 
-    this.$elm.classList.add('grid-listing');
+    this.$elm.classList.add('grid-listing__about');
```
nlisgo, line 14: are there styles that we can borrow from `grid-listing` too and this can be a variant?

discodavey: I think I have used all I can apart from the breakpoint. I will have another look this morning.

discodavey: I have created three mixins so these can be used between both grid-listing and grid-listing__about.

discodavey:
Closing as no longer needed.

PR #1050, opened 2021-01-07
Remove carousel item pattern and move examples to carousel

nlisgo:
As carousel item is not currently work or any of it's variants, I have moved all of the variants into carousel in the `organisms` section and have moved the `scss` and template and schema to the `carousel` template.

PR #1051, opened 2021-01-08
Add CarouselItem javascript behaviour

nlisgo:
This is a different approach to https://github.com/elifesciences/pattern-library/pull/1050

But this has the benefit of requiring no changes to php code in `patterns-php` or `journal`. It simply moves the little bit of javascript to get the template to be useable into it's own file `CarouselItem.js` and out of `Carousel.js`.


discodavey (approved):
All seems to work as production. No errors in JS either.

PR #1052, opened 2021-01-11
change javascript to target all collapsible sections for fragment handling

nlisgo:
This should fix an issue when the fragment is anchored to a figure supplement which is introduced to the article page with javascript.


test/fragmenthandler.spec.js
```diff
@@ -27,72 +27,72 @@ describe('A FragmentHandler Component', function () {
...
-        expect(fragmentHandler.getIdOfCollapsedSection('arbitraryElementId',
+        expect(fragmentHandler.getIdOfCollapsibleSection('arbitraryElementId',
                                                        docMock)).to.be.null;
       });
 
-      it('returns null if hash does not identify a fragment of or within any collapsed sections',
+      it('returns null if hash does not identify a fragment of or within any collapsible sections',
       function () {
 
-        let collapsedSectionIds = [ 'collapsedA', 'collapsedB', 'collapsedC' ];
```
discodavey, line 57: How many collapsed sections are we going to have. If there is more than three do we need to think of a better naming convention?

nlisgo: > How many collapsed sections are we going to have. If there is more than three do we need to think of a better naming convention?

This is a test fixture.

discodavey: Ok. I need to just test it locally.

discodavey: Just tested it locally. Looks ok. Anything you need me to test in particular to make 100% sure?

discodavey (approved):
Looks ok. Like my comment below says I have tested it locally and looks ok. Is there anything you want me to test in particular to make sure?

PR #1053, opened 2021-01-26
Switch mathjax v2 to v3

discodavey:


assets/js/components/Math.js
```diff
@@ -37,100 +37,35 @@ module.exports = class Math {
...
-
-    window.addEventListener('resize', resizeThrottler);
   }
 
   static load(doc) {
     let script = doc.createElement('script');
+    script.id = 'MathJax-script';
+    script.async = true;
     script.type = 'text/javascript';
-    script.addEventListener('load', Math.setupResizeHandler);
```
nlisgo, line 129: This could that has been removed was necessary presumably because the MathML need to be re-rendered each time the viewport changed. Does this need to be re-instated?

nlisgo:
I'm going to deploy these changes to an instance of journal

nlisgo:
We have an example in pages that should work but doesn't:

https://ui-patterns.elifesciences.org/patterns/04-pages-article--research/04-pages-article--research.html#equ1

When run locally it doesn't work on this feature branch.

nlisgo:
> We have an example in pages that should work but doesn't:
> 
> https://ui-patterns.elifesciences.org/patterns/04-pages-article--research/04-pages-article--research.html#equ1
> 
> When run locally it doesn't work on this feature branch.

This is working now.

PR #1056, opened 2021-02-05
Removing text-transform as not needed

discodavey:


source/_patterns/04-pages/article--research.mustache
```diff
@@ -60,11 +60,7 @@
             {{! May need to make the article section pattern more content flexible (include other patterns/not have header). New variants? }}
             {{#articleMeta}}
               <section class="article-section">
-
-                <div class="article-section__body">
```
NuclearRedeye, line 64: Was this change intentional? If so, does this also need to be made to any of the other pages examples to keep them in sync or not?

NuclearRedeye (approved):
Not tested locally, but changes look innocent enough to me. Minor comment about one change that might need to be replicated in other examples. This could be a followup task however if it is required.

source/_patterns/04-pages/article--research.mustache
discodavey: Yes it'sdone intentionally as it's not on production and if this class is in place you do not get the blue links

PR #1057, opened 2021-02-10
adding cc and downloads

discodavey:


NuclearRedeye (approved):
Only had a glance, the changes seem fine to me. What does this relate to however? Is this just fixing an inconsistency in the `pattern-library` pages or is it fixing another issue?

discodavey:
It's one of the pages that just show what an article page looks like from magazine but to be honest I do not think we need this page as any magazine links go to an article anyway. I think we could just remove this unless I am missing something.

PR #1060, opened 2021-02-17
Allow placeholder image option on about profile

nlisgo:
We need to allow the default image to be optional as we will not want it to appear in all cases.


discodavey:
@nlisgo Would there be any scenario for profile with no image without the placeholder?

nlisgo:
> @nlisgo Would there be any scenario for profile with no image without the placeholder?

The reviewing editors.

discodavey (approved):

discodavey:
> > @nlisgo Would there be any scenario for profile with no image without the placeholder?
> 
> The reviewing editors.

I spotted this after I sent my message.

PR #1061, opened 2021-02-22
Adding underline to blue links

discodavey:


NuclearRedeye (commented):
In a side by side comparison of current vs this, I notice...

- `reference-book-chapter` atom title has no underline, seems to be the same for other references with blue titles.
- in `listing-read-more`, the msa's don't seem to be links. They are links in other areas. I don't think this needs to be fixed here, as it's not a result of this change, but is this something that should be fixed or is this the intended behaviour?

Other than the references titles, this seems fine to me.

discodavey:
> In a side by side comparison of current vs this, I notice...
> 
>     * `reference-book-chapter` atom title has no underline, seems to be the same for other references with blue titles.
> 
>     * in `listing-read-more`, the msa's don't seem to be links. They are links in other areas. I don't think this needs to be fixed here, as it's not a result of this change, but is this something that should be fixed or is this the intended behaviour?
> 
> 
> Other than the references titles, this seems fine to me.

References title does not need to have an underline. The changes for the reference changes have not been pushed live as yet and are in another PR so we do not need to worry about this.

The listing read more is a known thing and I believe there is a ticket somewhere for the future to make these links as it's the same on production.

NuclearRedeye (approved):

PR #1062, opened 2021-03-02
fix: podcast duration test failing due to missing content

NuclearRedeye:
This fixes elifesciences/issues#6362


discodavey (approved):
Have tested and works ok

PR #1064, opened 2021-03-12
Amend font size for doi

discodavey:


discodavey:
Please check doi in references and in page research to compare.

NuclearRedeye (approved):

PR #1068, opened 2021-03-19
Non doi link styling amends

discodavey:


discodavey:
Sorry.. closed this but should be ok. It's to get the blue link to work on the article page

PR #1070, opened 2021-03-22
Add another class for excluding popup

discodavey:


discodavey:
This needs to be put on a temp PR so it can be tested before we go any further.

nlisgo (approved):

NuclearRedeye:
@discodavey Can this be merged?

discodavey:
We need to add a test for the new additional code.

PR #1071, opened 2021-04-14
Updating footer links to match production

discodavey:


NuclearRedeye (approved):
Changes seem innocent to me, although I've not cross checked them

PR #1072, opened 2021-05-06
moving credit overlay into header

discodavey:


NuclearRedeye:
I was just testing this, and I can't find any page to test it. I notice that on the Pages -> MSA item that there isn't a credit, should you consider adding a credit so that it is visible in the pattern?

discodavey:
I have not done anything on pages purely cause of the decision with patttern library and no point in updating them if we are not going to use them. 

The place you need to look is on http://localhost:8889/?p=viewall-organisms-content-headers

NuclearRedeye (approved):

PR #1076, opened 2021-05-27
Add padding to list css class

discodavey:


discodavey:
Will push this through to journal to be checked there. Examples need to be added on pattern library and will create a new ticket for this to be done.

discodavey:
We are adding padding-left as default and then removing it from when in a table footer.

NuclearRedeye (approved):

PR #1079, opened 2021-07-05
Change text for Hypothesis call to action buttons

discodavey:


discodavey:
This is just for the pattern library. It looks like that the change will need to be done in Patterns PHP doe Journal.

NuclearRedeye (approved):
@discodavey This is indeed somewhat on the confusing side. Based on what you say above, looks to me like the actual place the text is set is [here](https://github.com/elifesciences/patterns-php/blob/master/src/ViewModel/SpeechBubble.php#L47), but I am surprised that isn't just an atom in pattern-library.

discodavey:
I thought it was in patterns-php but will change both.

PR #1082, opened 2021-08-17
etoc redesign, remove form field, add button and styling

discodavey:


discodavey:
@NuclearRedeye @nlisgo Just thinking about this.. Will anything need to be taken out of Journal for the functionality of the form? If so then do we do it on this ticket or raise another issue?

NuclearRedeye:
@discodavey I'd suggest that any cleanup be done (unless trivial) as a followup to this PR if there is any like you say. You might also need to see if there are any tests that either need updating or adding.

discodavey:
@nlisgo Happy with what has been done. Have made one change which is padding for header text

PR #1083, opened 2021-09-01
Add styling for Cookiebot privacy focus removal

discodavey:
This is more of a hack as it's a third party app. Will not be able to test this until we get it into Journal as that is where the  Cookiebot code sits.


NuclearRedeye (approved):
Not tested, but looks OK to me

discodavey:
> Not tested, but looks OK to me

We can find out if it works when we get it through to testing on a temp PR in Journal

PR #1085, opened 2021-09-24
Adapt article section to accommodate classes and relatedLinks

nlisgo:
We want to offer the flexibility to present an article section within a padded box and with related links. I would rather adapt an existing template that could potentially be used in other places. No need to have anything that only works for editor evaluations.

@discodavey I need a review and feedback on the markup changes and new values available to the article section template and some help with the css.

This is what we want it to look like: https://projects.invisionapp.com/share/GC117EOOQ6DE#/screens/453719057_higher-Summary-With-Link

### DIFF
--- a/assets/js/components/HypothesisOpener.js
+++ b/assets/js/components/HypothesisOpener.js
@@ -284,7 +284,7 @@ module.exports = class HypothesisOpener {
   }
 
   static positionBySecondSection($elm, $contentContainer) {
-    const $firstSection = $contentContainer.querySelector('.article-section--first');
+    const $firstSection = $contentContainer.querySelector('.article-section--first:last-child');
     if ($firstSection) {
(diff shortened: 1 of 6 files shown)


discodavey:
@nlisgo Do we need to make the id dynamic as if we end up with two highlighted article sections then we will need to change the id?

nlisgo:
> @nlisgo Do we need to make the id dynamic as if we end up with two highlighted article sections then we will need to change the id?

The `id` for each section is set in journal and responsibility lies there to ensure the value is unique.

nlisgo:
Have decided it's best to place the related links within `article-section__body` so that the collapsible behaviour works. Doing this may mess with the styling a little. It's getting close though.

discodavey:
@nlisgo I have moved the related outside of hightlighted and should work ok.

PR #1086, opened 2021-10-06
Add checkboxes pattern

nlisgo:


chris-huggins:
Can we please reduce the space between the top heading ("I would like...") and the checkbox below by 12px/.75rem?

PR #1087, opened 2021-10-13
Sidebar navigation expanded authors list issue

discodavey:


NuclearRedeye:
Tested locally, it looks like it does resolve the problem. There is one tweak though that I think you should look into. I notice when it transitions from relative to fixed, there is some padding applied which seems a little too much to me as the bar jumps noticeably by a few pixels, likewise when it transitions the other way. It didn't do this before, so think you might need to tweak the styling/calculations a little if you can.

NuclearRedeye:
I was going to say, think you might've pushed those fixes onto the wrong branch

NuclearRedeye (approved):

discodavey:
Just run the tests and the second one seems to fail still.

Screenshot 2021-11-15 at 09 15 58

NuclearRedeye:
@discodavey work fine for me locally, and also on CI. Seems the issue is isolated to you, so you'll need to try and debug what is happening when you run those tests locally in your environment and assert what is different.

discodavey:
Looks like there are issues with the bin commands as @NuclearRedeye @will-byrne and myself have just found out while trying to get the tests to run locally on my machine.

Will need to create a ticket for these to be looked into at some point soon so that tests can be run without any issues.

Merging this into production

PR #1088, opened 2021-10-19
Nav bar link changes

discodavey:


discodavey:
This is for the pattern library changes with the CSS which will go through to Journal. The actual navigation changes will be done in Journal itself.

PR #1093, opened 2021-11-02
Fix IOS issue with checkboxes and refactor CSS

discodavey:


discodavey:
Initial fix for the checkboxes to be aligned correctly with label

NuclearRedeye (approved):

PR #1094, opened 2021-11-11
Initial content header framework build

discodavey:


discodavey:
This is the initial framework for the new content header. We need to integrate this fully so all the other content headers look ok.

assets/sass/patterns/organisms/content-header.scss
```diff
@@ -25,6 +24,49 @@
...
+  border-bottom: 1px solid $color-text-dividers;
+}
+
+.content-header-grid__side {
+  @include margin(24, "bottom");
+  @include padding(24, "bottom");
+  border-bottom: 1px solid $color-text-dividers;
+}
+
+@media only all and (min-width: 900px) {
```
NuclearRedeye, line 45: In other locations in this file, these @media definitions are nested under the respective class. Would it not make more sense to replicate that model? E.g..

```
.conter-header-grid {
  padding: 0, 24px;

  @media only all and (min-width: 900px)
    border-bottom: 1px solid $color-text-dividers;
    display: grid;
    grid-template-columns: repeat(12, 1fr);
    grid-gap: 24px;
    max-width: 1114px;
    padding: 0;
  }
```

NuclearRedeye, line 45: Also, wouldn't it be better to store the breakpoint value in a variable? e.g.

```
(min-width: $bkpt-size--medium)
```

source/_patterns/02-organisms/content-headers/content-header.mustache
```diff
@@ -5,6 +5,147 @@
     class="content-header {{#image}}content-header--image content-header--header{{/image}} {{^image}}wrapper {{#header.possible}}content-header--header{{/header.possible}}{{/image}} {{#socialMediaSharers}}content-header--has-social-media-sharers{{/socialMediaSharers}} {{#header.hasProfile}}content-header--has-profile{{/header.hasProfile}} clearfix"
```
NuclearRedeye, line 5: Any reason not to just add the `content-header-grid` class to the `header` element?

assets/sass/patterns/organisms/content-header.scss
```diff
@@ -25,6 +24,49 @@
...
+
+  .content-header-grid {
+    border-bottom: 1px solid $color-text-dividers;
+    display: grid;
+    grid-template-columns: repeat(12, 1fr);
+    grid-gap: 24px;
+    max-width: 1114px;
```
NuclearRedeye, line 52: Do we not have a variable that can be used for max width? I'd imagine that it's the value needs to be the same for a large number of sections of the site.

assets/sass/patterns/organisms/content-header.scss
```diff
@@ -25,6 +24,49 @@
 
 }
 
+// Content Header Grid Articles
+
+.content-header-grid {
+  padding: 0 24px;
```
NuclearRedeye, line 30: I see the `24px` value used a lot in these changes. Is there not a variable similar to `$content-header-padding` that can be used? If not, would it not be wise to add one?

NuclearRedeye (commented):
Some feedback from a look through the code. Most suggestions are around making better use of `sass` to avoid repetition and make things easier to maintain moving forwards.

assets/sass/patterns/organisms/content-header.scss
discodavey: Agree that this how it is currently done but will want to review this at some point in the future if the Journal survives

discodavey: Medium is 730. wide is 900 but yes we can do that as it makes sense.

source/_patterns/02-organisms/content-headers/content-header.mustache
discodavey: This is a conversation I will be having with Nathan and how it will integrate with the rest of the Content Headers as this exercise was just to get the new header in to start

assets/sass/patterns/organisms/content-header.scss
discodavey: No we don't have this I think we can create a variable for it. There is two of these now I have implemented it into content header. The other is in info bars.

discodavey: There is max-site-width which is 1114px. There is one other place that uses 1114px which is in infobars so will change this as well.

assets/sass/patterns/organisms/content-header.scss
NuclearRedeye: What do you mean by `if the Journal survives`?

assets/sass/patterns/organisms/content-header.scss
discodavey: I have set up $site-spacing-bottom in the variables sass file.

assets/sass/patterns/organisms/content-header.scss
discodavey: If it does not get put into maintenance mode.

PR #1095, opened 2021-11-24
fix: ios styling issue for checkboxes

discodavey:


discodavey:
I have made one small change to the checkbox. It works in Pattern Library but need to check with a temp pr in Journal to make sure all is ok.

NuclearRedeye (approved):

PR #1101, opened 2022-01-07
Journal article main content updates

discodavey:


discodavey:
This is purely for the article section and not to do with the content header itself.

NuclearRedeye (commented):
Tested locally, obviously not got the same keen eye as a designed but functionally it looks OK to me except for the buttons not functioning, tracked in #6966 & #6967, and the social menu not collapsing at mobile resolutions, #6969.

I've put some comments on the ticket which could be addressed if relevant, but this can be merged to the feature branch if needed, which I note that this PR is raised against `develop` rather than a content-headers specific feature branch. That needs to be corrected as we don't I think want these changes on develop at the moment.

PR #1102, opened 2022-01-12
Add Carbon Neutral logo to footer

discodavey:


NuclearRedeye (commented):
Changes look OK to me. There is a subtle discrepancy against the design, but to me it looks like the problem is the social icons are spaced different between the css and design file rather than the icon itself being the issue.

Screenshot 2022-01-12 at 08 47 39

There isn't anything in the code that stands out as an issue to me, hence we should get this merged so that we can create an ad-hoc journal instance for Chris to review next week.

will-byrne (approved):

PR #1104, opened 2022-01-21
Add table variations and fix bug for lists inside tables

discodavey:


assets/sass/patterns/atoms/list.scss
```diff
@@ -6,13 +6,6 @@
   padding-left: 0.75rem;
 }
 
-// Had to put this here as list.scss overrides table.scss
```
nlisgo, line 9: Does this no longer apply?

source/_patterns/00-atoms/components/list.mustache
```diff
@@ -17,3 +17,128 @@
     </ul>
 
 {{/isOrdered}}
+
+
+ <ul class="list">
```
nlisgo, line 22: This should be removed

source/_patterns/00-atoms/components/table.mustache
```diff
@@ -22,5 +22,4 @@
     </ol>
 
   {{/hasFootnotes}}
-
-</div>
+</table>
```
nlisgo, line 25: There is no opening `</table>` in this template.

nlisgo: I think this should be reverted.

nlisgo (changes requested):
I've added comments.

assets/sass/patterns/atoms/list.scss
discodavey: Have you got the latest version as I removed this on Friday?

nlisgo (approved):

PR #1110, opened 2022-02-24
Introduce modal window

nlisgo:
Also minor changes to doi to support presentation without a link.

This should be safe to merge into `develop`. I have reintroduce js to handle the modal remaining open on resize unless `smallDeviceOnly` is set and the resize takes the window above 480.

I would prefer it if we introduce a test for `Modal.js` but we can add that as a followup.


discodavey:
@nlisgo Have you left out the buttons for a reason as they are missing from both the citations and share modals?

discodavey:
My own opinion is that the modal window should stay visible when resized. My theory behind this is if the user is on their mobile and turns to landscape or vice versa then they will look for the modal but it will be gone which will confuse and possibly annoy them.

assets/sass/patterns/atoms/modal-window.scss
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+.modal-content__body {
+  display: grid;
+  grid-column: 1/3;
+
+  .button {
+    justify-self: center;
+    margin-bottom: 24px;
```
discodavey, line 54: I did changes for the css in 1107 but these have not been brought over.

nlisgo: Discussed this. Will reintroduce when we bring together in content-header pr.

discodavey (approved):
All looks good and works well

nlisgo:
I will follow up with a test from `Modal.js`.

PR #1111, opened 2022-02-24
Introduce button clipboard and button secondary

nlisgo:
This should contain only changes for `button-clipboard` and `button-secondary`. This should be safe to merge into `develop` branch.

~I will try and add work here to check if `navigator.clipboard` api is available and hide the button otherwise.~

I have added in a check to see if clipboard is supported. I also offer a fallback if `navigator.clipboard` isn't supported by the deprecated `document.execCommand('copy')` is. If neither of those two approaches work then the clipboard button is removed.


nlisgo:
I'm assuming approval from @discodavey based upon our discussions.

PR #1107, opened 2022-02-08
Share popup modal separation

discodavey:

PR #1113, opened 2022-02-25
New content headers - article pages - phase 1

nlisgo:
This combines the work in both of these PR's:

- https://github.com/elifesciences/pattern-library/pull/1094
- https://github.com/elifesciences/pattern-library/pull/1107


discodavey:
@nlisgo Need the modals in this PR so I can make the changes that Chris has asked for.

source/_patterns/00-atoms/components/social-media-sharers-journal.mustache
```diff
@@ -0,0 +1,7 @@
+<ul class="social-media-sharers">
+  <li><a class="social-media-sharer email" href="{{emailURL}}" target="_blank" rel="noopener noreferrer" aria-label="Share by Email"><span class="visuallyhidden">Email</span></a></li>
```
nlisgo, line 2: This should be `emailUrl` to be consistent with the schema.

assets/sass/patterns/molecules/article-download-links-list.scss
```diff
@@ -76,22 +77,30 @@
...
-    right: 16px;
+    right: 45px;
   }
 
   @media only all and (min-width: #{get-rem-from-px($bkpt-content-header--extra-wide)}em) {
     right: 0;
   }
 }
+
+.article-download-links-list__group--exclude {
```
nlisgo, line 102: @discodavey why was it necessary to add this class could we not have targeted an existing class. Introduced here: https://github.com/elifesciences/pattern-library/commit/b8c8c0c5f09de1ab6173fc6a0d72e8e06cbbdcba

discodavey: It would have affected other sections in the page that did not need the changes. I did not want to use the class article-download-links-list__group--js-exclude as it's just for JS only.

nlisgo: It's safe to just target `.article-download-links-list__intro` in all cases.

PR #1115, opened 2022-03-02
Initial one column grid template build

discodavey:


discodavey:
@nlisgo Can you have a look at this please. I have left the majority of the template intact but reduced the json and yaml files to simplify as discussed.

PR #1117, opened 2022-03-21
Upgrade node

will-byrne:


.nvmrc
```diff
@@ -1 +1 @@
-lts/boron
\ No newline at end of file
+lts/gallium
```
nlisgo, line 1: nitpick: newline

nlisgo:
I've tested all steps in the current readme. All worked well with some adjustments to Makefile and readme.

There was one test that failed for me:

```
it('dispatches a chapterChanged event on the supplied element',
```

Will create a ticket to add a `make` step to allow individual tests to be run but that is not blocking.

Biggest concern is the length of time it takes when running `make watch` and re-rendering the assets. A potential improvement would be to only render the sass if changes detected there rather than re-rendering all assets if any kind of asset changes.

Great work.

We will adapt the Jenkins pipelines as a followup to this work so will maintain the docker compose steps until then.

PR #1122, opened 2022-04-06
Remove color from reference__title

nlisgo:


nlisgo:
> Works well in the research pages on the pattern library. Happy to approve but we will need to run a PR on journal to check fully.

https://pr-1599--journal.elifesciences.org/

PR #1123, opened 2022-04-07
Allow PMC check before displaying ReadCube link

nlisgo:
@discodavey I'd lack to sanity check this with you before I progress on it further and introduce a test.

I think the main thing we need to agree on is what is it acceptable to have in the DOM before JS performs the PMC check. Only when that check is acceptable will we introduce the link.

### DIFF
--- a/source/_patterns/01-molecules/components/article-download-links-list.mustache
+++ b/source/_patterns/01-molecules/components/article-download-links-list.mustache
@@ -18,10 +18,15 @@
             {{/secondary}}
 
             {{#primary}}
-              <a href="{{url}}"
-                       class="article-download-links-list__link"
-                       {{#attributes}}data-{{key}}{{#value}}="{{value}}"{{/value}}{{/attributes}}
-                    >{{{name}}}</a>
+              {{#checkPMC}}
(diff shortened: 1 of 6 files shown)


discodavey (commented):
When you turn off JS there are two bullets showing after Mendeley that are not showing any links.

source/_patterns/01-molecules/components/article-download-links-list.mustache
```diff
@@ -18,10 +18,15 @@
             {{/secondary}}
 
             {{#primary}}
-              <a href="{{url}}"
-                       class="article-download-links-list__link"
-                       {{#attributes}}data-{{key}}{{#value}}="{{value}}"{{/value}}{{/attributes}}
-                    >{{{name}}}</a>
+              {{#checkPMC}}
+                <span data-behaviour="CheckPMC" data-check-available="{{checkPMC}}" data-display-on-available="<a href='{{url}}' class='article-download-links-list__link'{{#attributes}} data-{{key}}{{#value}}="{{value}}"{{/value}}{{/attributes}}>{{{name}}}</a>"></span>
```
discodavey, line 22: Missing greater than for span tag

PR #1127, opened 2022-05-04
Extract authors and institutions into separate template

nlisgo:


nlisgo:
Here is a demo instance of journal with this work incorporated: https://pr-1609--journal.elifesciences.org

discodavey (changes requested):
Looks good but needs some styling in the authors molecule for padding between authors and institutes.

nlisgo:
> Looks good but needs some styling in the authors molecule for padding between authors and institutes.

@discodavey I think the padding is only present on the content header redesign and this PR is intended to maintain the same presentation as is currently implemented on the live site.

discodavey (approved):
All looks good from testing

PR #1132, opened 2022-05-18
Make title an optional property of article-section

nlisgo:


discodavey (approved):
All looks good from testing

PR #1136, opened 2022-05-24
styling counters for views and citations

discodavey:


nlisgo (approved):
Don't love setting classes in the json but will indicate the need to refactor once I integrate with journal and perhaps we can address this once we have migrated all content to a new form of contextual data.

PR #1139, opened 2022-06-07
Add comment for bug in firefox

discodavey:


nlisgo:
This is a followup to the fix for: https://github.com/elifesciences/issues/issues/7451

PR #1145, opened 2022-08-11
Msa label in listing and article page

alexdinug:


Maelplaine:
@alexdinug Can you remove me as a PR reviewer? I will only review the UI with Chris when it is available in a staging environment

PR #1152, opened 2022-08-24
Removing as causing weird issues with links

discodavey:


discodavey:
I am pushing this through to see what it looks like on a journal temp PR as I think this may cause other issues. This looks ok in Pattern Library at the moment but if there are issues then we will need to add them to Pattern Library for future reference.

nlisgo:
I'm going to deploy this through to a demo instance of journal.

discodavey:
@scottaubrey or @nlisgo Can you please review this code please as I need to get this live this morning.

discodavey:
Going to merge this into master.

PR #1155, opened 2022-08-24
Updated README.md - added windows dependencies and reorganized the in…

achirazi:
Added windows dependencies and reorganized the installation guide.


scottaubrey (commented):
For what it's worth, the main build fails on composer v2 (similar to journal), and it might be worth documenting that they specifically require composer v1 for the main commands to work.

Otherwise, I was able to get up and running with the docker-compose v1 (and a smattering of platform-overrides for arm). 

Could the Dependencies sections be condensed to just a single list for all platforms, with a special mention for windows WSL?

nlisgo (approved):

PR #1157, opened 2022-08-25
Remove carousel files and logic

viktoria-mahmud:


assets/js/components/Highlights.js
```diff
@@ -127,7 +127,7 @@ module.exports = class Highlights {
     this.margin = this.calculateMargin() * 2;
 
     // Update some things.
-    this.carouselWidth = window.getComputedStyle(this.$elm).width.match(/([0-9\.]+)px/)[1];
+    this.contentWidth = window.getComputedStyle(this.$elm).width.match(/([0-9\.]+)px/)[1];
```
nlisgo, line 130: Unfortunately we don't have tests on Highlights.js but I have tried it out and this works as expected. We could create a ticket add a test to Highlights.js as a followup

nlisgo (approved):

PR #1158, opened 2022-08-29
Update all examples of author line

Cosmin-Daniel:
Update all examples of author line with "[first author], [second author] ... [last author]"


nlisgo (approved):
I'm happy with this. Thanks for this, happy for this to be merged in.

discodavey:
Slightly confused with this as the teasers in the home.json file use the class "teaser__secondary_info" where as in highlights it is using "author-line"

As said in another ticket which I think is highlights we should be using Authors with the same css classes using lists.

I will create a new ticket for this work to be done.

PR #1162, opened 2022-09-12
fix-spacing for licence icons

discodavey:
fixes https://github.com/elifesciences/issues/issues/7547


scottaubrey (approved):
Checked on the Preview env - looks good to me. Do you need anything else from me @discodavey ?

nlisgo (approved):

PR #1164, opened 2022-09-14
Reformat status and date in teaser

Cosmin-Daniel:


Cosmin-Daniel:
[The issue can be seen here](https://github.com/elifesciences/issues/issues/7715)

QueenKraken:
Reviewed and happy. thanks Cosmin.

nlisgo:
@viktoria-mahmud are you able to do a review on this?

nlisgo:
At a quick glance we have an issue that we can only see the styles impact the teaser in the pages section with the `listing-column-main` class in play. We should be able to demonstrate that we can present the teaser as we want in isolation and not only if it happens to be within a particular section of a page.

We should consider how the `rootClasses` parameter is being used. The main style should be the one in the designs and we can override it for secondary and grid style teasers.

Happy to discuss any of this tomorrow morning.

Cosmin-Daniel:
> @Cosmin-Daniel can we close as this PR combines that work:
> 
> * [Include status in item listed #1150](https://github.com/elifesciences/pattern-library/pull/1150)
> * [Include date in item listed #1153](https://github.com/elifesciences/pattern-library/pull/1153)

I have closed #1150 and #1153 PRs.

viktoria-mahmud:
The review from my side is done.

discodavey (approved):
Looks ok to me

discodavey (approved):
Happy with the latest changes

source/_patterns/02-organisms/components/listing-teasers.mustache
```diff
@@ -1,4 +1,4 @@
-<div{{#highlights}} class="highlights"{{/highlights}}>
+<div{{#highlights}} class="highlights"{{/highlights}}{{^highlights}} class="teasers"{{/highlights}}>
```
nlisgo, line 1: This could be simplified but not necessary:

```
<div class="{{#highlights}}highlights{{/highlights}}{{^highlights}}teasers{{/highlights}}">
```

nlisgo:
@chris-huggins can you confirm whether you are happy for the formats (HTML, PDF) can be removed from the teaser in this release or should we keep them in for now?

Example with formats: https://s3.amazonaws.com/ci-pattern-library/1164/patterns/01-molecules-teasers-teaser-70-main--big-status-and-date/01-molecules-teasers-teaser-70-main--big-status-and-date.html

chris-huggins:
@nlisgo Please keep the formats. I think you might have missed the conversation last week where we decided to reinstate them as we will be keeping POA for the time being. 

@Cosmin-Daniel Is it possible for the date to wrap under the status as a block instead of leaving the Month on the line above?

Cosmin-Daniel:
> @Cosmin-Daniel Is it possible for the date to wrap under the status as a block instead of leaving the Month on the line above?

Yes, it is possible, but the thing is that now the dot between the status and the date is used as a ":before" for date and would not look good when the date is on a new line/row. @discodavey Would be ok for you to use that dot as an ":after" for the status instead of ":before" for date?

discodavey:
@Cosmin-Daniel Can you make the date drop under the type please when the width goes below 480px. This means the the bullet will need to disappear at the same time as well as the date being in one line and not wrapping. You will have to either adjust the grid widths or use flexbox for the teaser__footer class.

Also there are no examples in teasers with the updated next to the date which will be need to be done in the molecules/teasers section of the pattern library.

Does this make sense.

There maybe a few more smaller changes tomorrow morning after myself and Nathan have reviewed this together.

Any questions then please ask

Cosmin-Daniel:
> @Cosmin-Daniel Can you make the date drop under the type please when the width goes below 480px. This means the the bullet will need to disappear at the same time as well as the date being in one line and not wrapping. You will have to either adjust the grid widths or use flexbox for the teaser__footer class.

@discodavey Now the date goes under the status as a block and the bullet disappears below 480px. My suggestion is to go for the next breakpoint (730px) because there are still screens where the month is beside the status.

> 
> Also there are no examples in teasers with the updated next to the date which will be need to be done in the molecules/teasers section of the pattern library.

I am not sure I understood that completely. The new molecule is there, called "teaser-main--big-status-and-date" (https://s3.amazonaws.com/ci-pattern-library/1164/?p=molecules-teaser-main--big-status-and-date). Please let me know if there is anything else I need to do.

assets/sass/patterns/atoms/date.scss
```diff
@@ -6,6 +6,9 @@
 
   .teaser--main & {
     @include font-size(14);
+    @media only all and (max-width: 480px) {
```
discodavey, line 9: Can we use the variable for the width please.

assets/sass/patterns/molecules/meta.scss
```diff
@@ -29,6 +29,12 @@
     content: "\00a0\2219\00a0";
     font-weight: bold;
   }
+
+  @media only all and (max-width: 480px) {
```
discodavey, line 33: Can we use the variable for the width please

assets/sass/patterns/atoms/date.scss
Cosmin-Daniel: Ok. Can you confirm we are going with the 480px breakpoint?

discodavey:
The date needs to drop under on all variations where there is a type and date. Agree with the 730 breakpoint instead of 480px.

assets/sass/patterns/atoms/date.scss
discodavey: @Cosmin-Daniel See my comment at the bottom of the PR

Cosmin-Daniel: A variable is used now for the width.

assets/sass/patterns/molecules/meta.scss
Cosmin-Daniel: A variable is used now for the width.

discodavey:
The wrapping looks all ok. This just now needs to happen across all variants and we also need a variant with the "updated" in it next to the date.

discodavey:
Just spoke with @chris-huggins all ok. Happy with how it is now.

viktoria-mahmud (approved):

nlisgo:
@Cosmin-Daniel and @discodavey are you happy with the last couple of commits? If so, we can merge.

PR #1149, opened 2022-08-18
Add initial hero banner

viktoria-mahmud:

PR #1167, opened 2022-09-28
Update MSA picker + remove strapline

alexdinug:


viktoria-mahmud (approved):

source/_patterns/01-molecules/components/section-listing.mustache
```diff
@@ -2,7 +2,7 @@
   {{#listHeading}}
     {{> atoms-list-heading}}
   {{/listHeading}}
-  <ol class="section-listing {{#singleLine}}section-listing--single-line{{/singleLine}}" {{#labelledBy}}aria-labelledby="{{labelledBy}}"{{/labelledBy}}>
```
discodavey, line 5: What was the reason for removing the labelledBy aria tag

alexdinug: The reason I deleted that line was that it was no longer necessary. I asked before if it is okay to delete what is no longer necessary.

discodavey (approved):
Looks ok to me

source/_patterns/04-pages/home.mustache
```diff
@@ -15,12 +15,6 @@
         </div>
       {{/hero-banner}}
 
-      {{#leadParas}}
-        <div class="wrapper">
-          {{> molecules-lead-paras}}
```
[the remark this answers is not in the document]
nlisgo: Ignore this. We are removing as part of this PR.

nlisgo (changes requested):
This is not ready to be merged now but could be with a little adjustment.

We need to ensure that the new MSA picker works well with the carousel and once @chris-huggins is happy we can release it ahead of the hero and highlight feature flag being lifted.

We need reintroduce carousel to a variant of homepage. Once we have that we can show @chris-huggins and agree on what spacing to introduce between the carousel and MSA.

nlisgo:
Here is where we removed the carousel: https://github.com/elifesciences/pattern-library/pull/1149/files#diff-9d9418f9fc9ecbd98eedf828fbafca8f4ef3ec42b4be6b9d9c2a615b185e08d4

Should be easy to create a variant of the home page with carousel. Here to help if you need.

nlisgo:
I've introduce a variant of the home page with the carousel. After a brief discussion with @chris-huggins we do need to consider how the style of the msa picker affects a similar element on other pages.

Let's schedule a call at the earliest convenience between @alexdinug @chris-huggins and myself.

nlisgo (changes requested):
We need to remove the `.idea` folder and the `install_nvm.sh` file

PR #1168, opened 2022-09-30
Change summary field to be optional for Highlights

viktoria-mahmud:
Create examples with and without summary for collections


viktoria-mahmud:
@nlisgo, do I need to create more examples without summary field? Please, add your comments

PR #1173, opened 2022-10-06
Add article page banner

viktoria-mahmud:


source/_patterns/04-pages/article--research.mustache
```diff
@@ -7,6 +7,14 @@
     </div>
     <main role="main" class="main blg-margin-bottom-large" id="maincontent">
 
+      {{#dismissible}}
+        <div class="wrapper">
+          <div class="content-container-grid">
```
nlisgo, line 12: The challenge accepting this in its current form is that we need to wrap the dismissible info bar in the `content-container-grid` so that it can be presented as we want but no other info bars work well wrapped in the same elements.

The info bar should be presented as you expected them to  be in the atoms preview, not only when viewed in a page with markup that doesn't work for all cases.

viktoria-mahmud: The additional wrappers are removed, all styles are changed for "info-bar--dismissible" class

nlisgo:
Thought it would be useful to display a page with multiple info bars: https://s3.amazonaws.com/ci-pattern-library/1173/patterns/04-pages-article--research-with-info-bars/04-pages-article--research-with-info-bars.html

@chris-huggins are you happy? You can also see the preview dismissible info bar in the atoms also: https://s3.amazonaws.com/ci-pattern-library/1173/patterns/00-atoms-components-info-bar-dismissible/00-atoms-components-info-bar-dismissible.html

nlisgo:
We should compare this approach with the feature flagged dismissible info bar on a corrected article page: 
https://preview--journal.elifesciences.org/articles/64509

chris-huggins:
> We should compare this approach with the feature flagged dismissible info bar on a corrected article page: https://preview--journal.elifesciences.org/articles/64509

Hopefully the 24px padding on the top of the bar I mentioned above would fix this? I'm not sure why there's no padding here, but 12px in pattern library. I don't mind how it's done (if it needs to be via pattern library or not) but would really like to see 24px between the info bar and dismissible bar before we deploy.

viktoria-mahmud:
> 

@chris-huggins, I added the changes you requested: 

.info-bar--dismissible now has 24px padding top
.info-bar--dismissible .info-bar__container has padding 14px 15px 9px for all screen sizes

chris-huggins:
Thanks @viktoria-mahmud This looks good to me.

nlisgo (approved):

PR #1175, opened 2022-10-06
Reduce bottom margin on site header for wide screens

nlisgo:


nlisgo:
@chris-huggins as discussed we are reducing the margin on the bottom of the site header for wider screens. This was because of the balance with the carousel on the home page but also because of how the site header sits with other banners. Once we go live with the hero banner we may wish to add a few px of margin on the hero banner to reflect the design but leave the margin as is elsewhere.

PR #1178, opened 2022-10-07
Add tabs to homepage and MSA pages on mobile

LoredanaChiuta:


nlisgo:
I will review this on Monday.

nlisgo:
Apologies. It's Wednesday and still no review. I think this review is going to have to be interactive. I need to initially talk through the changes with you.

I like the idea of repurposing the code used for tabs on article pages for the home page but if we agree this is the correct approach then a refactor of the existing behaviour to the article pages should proceed an attempt to expand the behaviour to other areas so that should become a 2 phase deployment. First like for like and then second introducing the new behaviour to the home page.

source/_patterns/01-molecules/navigation/view-selector.mustache
```diff
@@ -1,11 +1,11 @@
-<div class="view-selector{{#figureUrl}} view-selector--has-figures{{/figureUrl}}" data-behaviour="ViewSelector"{{#sideBySideUrl}} data-side-by-side-link="{{sideBySideUrl}}"{{/sideBySideUrl}}>
+<div class="view-selector{{#figureUrl}} view-selector--has-figures{{/figureUrl}}{{#isTabSelector}} button--switch-selector{{/isTabSelector}}" data-behaviour="ViewSelector"{{#sideBySideUrl}} data-side-by-side-link="{{sideBySideUrl}}"{{/sideBySideUrl}}>
   <ul class="view-selector__list">
     <li class="view-selector__list-item view-selector__list-item--article{{^figureIsActive}} view-selector__list-item--active{{/figureIsActive}}">
-      <a href="{{articleUrl}}" class="view-selector__link view-selector__link--article"><span>Article</span></a>
+      <a href="{{articleUrl}}" class="view-selector__link view-selector__link--article{{#isTabSelector}} primary-selector{{/isTabSelector}}"><span>Article</span></a>
```
discodavey, line 4: Please remove the spans

source/_patterns/04-pages/msa.mustache
```diff
@@ -24,21 +24,25 @@
...
-    <div class="wrapper">
+    <div class="wrapper wrapper--listing wrapper--listing-selector">
 
       <div class="grid">
 
         <div class="listing-column-container">
 
-          <div class="grid__item large--eight-twelfths listing-column-main">
+          {{#viewSelector}}
+            {{> molecules-view-selector }}
```
discodavey, line 34: Is the view selector supposed to be on the homepage?

viktoria-mahmud: Yes, it should be on Home and MSA

source/_patterns/01-molecules/navigation/view-selector.mustache
```diff
@@ -1,11 +1,11 @@
...
     </li>
     {{#figureUrl}}
       <li class="view-selector__list-item view-selector__list-item--figures{{#figureIsActive}} view-selector__list-item--active{{/figureIsActive}}">
-        <a href="{{figureUrl}}" class="view-selector__link view-selector__link--figures"><span>Figures and data</span></a>
+        <a href="{{figureUrl}}" class="view-selector__link view-selector__link--figures{{#isTabSelector}} secondary-selector{{/isTabSelector}}"><span>Figures and data</span></a>
```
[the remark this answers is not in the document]
viktoria-mahmud: It is done in another PR - https://github.com/elifesciences/pattern-library/pull/1199

source/_patterns/01-molecules/navigation/view-selector.mustache
viktoria-mahmud: It is done in another PR - https://github.com/elifesciences/pattern-library/pull/1199

source/_patterns/04-pages/msa.mustache
```diff
@@ -24,21 +24,25 @@
...
+          {{/viewSelector}}
+
+          <div class="grid__item large--eight-twelfths listing-column-main primary-column" id="primaryListing">
```
Cosmin-Daniel, line 37: The "one-whole" class is needed here too, as in home.mustache.

source/_patterns/04-pages/msa.mustache
```diff
@@ -24,21 +24,25 @@
...
 
             {{#mainTeasers}}
               {{> organisms-listing-teasers}}
             {{/mainTeasers}}
 
           </div>
 
-          <div class="grid__item large--four-twelfths listing-column-secondary">
+          <div class="grid__item large--four-twelfths listing-column-secondary secondary-column" id="secondaryListing">
```
Cosmin-Daniel, line 45: Here too, the "one-whole" class is missing.

source/_patterns/04-pages/msa.mustache
viktoria-mahmud: Done, also "large--eight-twelfths" is changed to "x-large--eight-twelfths"

source/_patterns/04-pages/msa.mustache
viktoria-mahmud: Done, also "large--eight-twelfths" is changed to "x-large--eight-twelfths"

source/_patterns/04-pages/home.mustache
```diff
@@ -27,17 +27,22 @@
        </div>
        {{/highlight}}
 
-      <div class="wrapper wrapper--listing">
+      <div class="wrapper wrapper--listing wrapper--listing-selector">
```
nlisgo, line 30: I don't think there is any need to introduce the wrapper--listing-selector class here. It's not needed for anything. The view-selector has all the info in the dom it needs.

source/_patterns/04-pages/msa.mustache
```diff
@@ -24,21 +24,25 @@
       </div>
 
 
-    <div class="wrapper">
+    <div class="wrapper wrapper--listing wrapper--listing-selector">
```
nlisgo, line 27: There is no need for the wrapper--listing or wrapper--listing selector classes.

chris-huggins:
@nlisgo I hope the below keeps this feature reasonably simple, but appreciate your thoughts on any issues that could be caused or if there might be a simpler approach?

> Do we need the ability to alter the url when we toggle between the view selector options?

I don't think it's necessary. Unlike the view selector on article pages, I see this is as just a JS enhancement that makes switching between two columns of content easier when reduced to one column on smaller screens. The tabs are intended to increase visibility and make access easier for the second column, which is easily missed below 1200px if it sits below the "load more" button for research.

> What is the expectation for non-js behaviour?

I think it is fine to just wrap/stack the magazine/highlights column under the latest research if JS is not enabled (essentially the existing behaviour below 1200px)

source/_patterns/04-pages/home.mustache
viktoria-mahmud: The class "wrapper--listing-selector" is not needed and I will remove it but "wrapper--listing" is adding aditional padding-top to the component and if we the js is disabled on the browser, we will hide the tabs and there won't be enough space between the sections, so for me it is better to reuse that class

source/_patterns/04-pages/msa.mustache
viktoria-mahmud: > Do we need the ability to alter the url when we toggle between the view selector options?

I got the impression by our conversation earlier this week that we need to keep the url logic. Also I can see that on prod it is used to jump on a parent section when the first tab is clicked (see the image below). Please correct me if I'm wrong, I will be happy if you advise me.

image

nlisgo:
I will have to review this tomorrow morning.

source/_patterns/01-molecules/navigation/view-selector.mustache
```diff
@@ -1,29 +1,40 @@
...
-          <a href="{{secondaryUrl}}" class="view-selector__link view-selector__link--secondary">{{secondaryLabel}}</a>
+        <span class="view-selector__link">{{secondaryLabel}}</span>
       </li>
-    {{/secondaryUrl}}
+    {{/isTabSelector}}
 
-    {{#otherLinks}}
-      <li class="view-selector__list-item">
-        <a href="{{url}}" class="view-selector__link"><span>{{{name}}}</span></a>
+    {{^isTabSelector}}
```
viktoria-mahmud, line 12: Everything inside {{^isTabSelector}} wapper is not changed, the old logic is moved there without updates, the only new logic is inside {{#isTabSelector}}

source/_patterns/01-molecules/navigation/view-selector~with-tabs.json
```diff
@@ -1,6 +1,4 @@
 {
-  "primaryUrl": "#primaryListing",
```
nlisgo: There is a lot of refactoring in order to avoid presenting the view-selector with links if it is a tab selector. My view is that the links still make sense. I know we hide them for non-js but semantically it still works to have the href of the link to reference the fragment id of the section that is being shown on click. 

Preserving the links will avoid the refactor of the yaml, I presume. Let's discuss tomorrow.

viktoria-mahmud:
@nlisgo, I am fine with these changes.

nlisgo (approved):
Let's merge this in on Monday. We need a little bit of work on patterns-php to allow the new parameter.

PR #1189, opened 2022-10-18
Announcement info bar

viktoria-mahmud:


chris-huggins:
Looks good to me, thanks @viktoria-mahmud

nlisgo:
@chris-huggins do you envisage that info bars of type announcement will always appear in the main content above anything else? Or may we want to view announcements in the standard info bar location which is underneath the site header?

source/_patterns/04-pages/about-elife.mustache
```diff
@@ -18,13 +18,17 @@
...
-
+     
       <div class="wrapper u-pad-top">
         <div class="listing-column-container">
           <div class="grid">
-
+            
           <div class="{{> pages-pl-only-grid-content-page-main-col-css-classes}} u-pad-top">
 
+            {{#infoBar}}
```
nlisgo, line 28: Maybe change this to `{{infoBarAnnouncement}}`?

viktoria-mahmud:
> @chris-huggins do you envisage that info bars of type announcement will always appear in the main content above anything else? Or may we want to view announcements in the standard info bar location which is underneath the site header?

I had a chat with Chris reagrding the announcement banner position, we will have the bar only under the manu.

image

nlisgo:
the logic I will follow for implementation of this info bar is that if it is of type announcement then it will appear in the main content section otherwise. If this needs to be different in future we can create a followup ticket to allow an infobar declaration to specify the position of the display.

PR #1190, opened 2022-10-20
New inside elife example in Pattern Library

LoredanaChiuta:


nlisgo:
@LoredanaChiuta can you ask for a review from @QueenKraken or @chris-huggins and then a colleague to review the PR.

LoredanaChiuta:
@nlisgo I have already done an internal review with Cosmin and Viktoria. After that I sent a message to Ash for the review. I have now also sent it to Chris if he can do a review.

chris-huggins:
Hi @LoredanaChiuta This is looking great, thank you. There are a few minor issues but otherwise it's very close.

1. There should ideally be more space between the date and the OA/CC-BY icons below 480px. Please try to match the space above 480px. But be aware of when the icons wrap onto a lower line on other article types. A margin after the date might be best.

2. Also there's more padding on the content of the article than needed - both at top/bottom and sides when above 1200px. I think it's because of the `<div>` with classes that are no longer needed (`.grid__item
one-whole
content-container
  large--ten-twelfths
  push--large--one-twelfth
  x-large--eight-twelfths
    push--x-large--two-twelfths
 u-pad-top`). I can't see it on Inside eLife posts on journal, and it seems to be causing the content to not go full width above 1200px.
 
3. It would be great to have more whitespace at the end of the main content of the article though. An additional 48px would be a good improvement I think. At the moment there isn't much space between the article content and the line above the text "Be the first to read new articles from eLife". 

Thanks

image

nlisgo:
> @nlisgo I have already done an internal review with Cosmin and Viktoria. After that I sent a message to Ash for the review. I have now also sent it to Chris if he can do a review. Now I'm waiting for their review.

@Cosmin-Daniel @viktoria-mahmud when you review and are happy with the state of a PR please, if appropriate, indicate by approving through github.

viktoria-mahmud (approved):

source/_patterns/04-pages/inside-elife-article.mustache
```diff
@@ -20,9 +20,9 @@
 
       <div class="wrapper wrapper--content">
 
-        <div class="grid">
+        <div class="content-container-grid content-header__one-column">
```
nlisgo, line 23: No need for the content-header__one_column class here!

We need to borrow more from the digest template. I think a short pairing session could get this into the shape it needs to be. Let's try tomorrow morning or Friday!

nlisgo (approved):

source/_patterns/04-pages/inside-elife-article.mustache
nlisgo: We should refer to:
- https://github.com/elifesciences/pattern-library/pull/1140/files
- https://github.com/elifesciences/journal/pull/1629/files

It may be that we need to adjust the logic for digests so that we can serve digests and inside-elife with a common logic.

chris-huggins:
@nlisgo @QueenKraken This PR looks good to me in pattern library.

QueenKraken:
@chris-huggins thanks Chris, you were marked as a reviewer in the PR, I think that means you need to to tick it off here in the PR?

chris-huggins:
@QueenKraken Ok, just done.

PR #1191, opened 2022-10-20
Remove carousel followup

nlisgo:
I wasn't happy with the changes to Highlights.js. listings with teasers as highlights is a carousel so we should restore the mention of carousel and this reflects in the use of carouselItem for listing teaser items. I did find that the use of carouselItem in meta and meta-journal is no longer needed.


nlisgo:
@viktoria-mahmud please check this over post-merge but I'm not happy leaving pattern-library as is so with a decision to revert the last commit or add this I'm going to add this. We can address any feedback you might have as a followup!

PR #1194, opened 2022-10-24
Remove span for content headers

viktoria-mahmud:


discodavey:
Looks ok from what I can see. This needs to be done on teasers as well. Will get a ticket created for that as well.

nlisgo:
@discodavey have you completed review on this?

discodavey:
Not looked at this in local but was there any styles in the class 'content-header__subject' and if so where have they been put as there are no changes to the css?

viktoria-mahmud:
> Not looked at this in local but was there any styles in the class 'content-header__subject' and if so where have they been put as there are no changes to the css?

The class "content-header__subject" was added on the span element and it is removed from the project now, most of the style is moved to the parent element "content-header__subject_link".

discodavey (approved):

nlisgo (approved):

nlisgo:
:ship:

PR #1198, opened 2022-11-03
Add research article with aside element

viktoria-mahmud:


assets/sass/patterns/molecules/article-section.scss
```diff
@@ -373,6 +373,60 @@ $arrow-size: 5;
   }
 }
 
+.section-wrapper {
+  @media only all and (min-width: #{get-rem-from-px($bkpt-site--small)}em) {
+    display: grid;
```
discodavey, line 378: As seeing as we are starting to use this in a few places we could make a mixin with lines 378,379 and 380

viktoria-mahmud: Good idea, thanks, Dave! The mixin grid is created in order to reuse the logic for grid display. The files info-bars.scss and content-header-journal.scss are not related to this PR but they can use the mixin too.

assets/sass/patterns/molecules/article-section.scss
```diff
@@ -373,6 +373,62 @@ $arrow-size: 5;
   }
 }
 
+.section-wrapper {
```
Cosmin-Daniel, line 376: A max-width and margin: 0 auto; may be needed here, as in the .content-header.

viktoria-mahmud: The class "section-wrapper" is updated with these styles, the content header will be handled in another task, now I am reusing an existing variant just for visual presentation

Cosmin-Daniel (approved):
Looks good to me.

source/_patterns/04-pages/article--research-aside.mustache
```diff
@@ -0,0 +1,91 @@
+<div class="global-wrapper" data-item-type="research-article">
```
viktoria-mahmud, line 1: After a call with Nathan, I created a new page for aricle research pages with aside element, the original page for article research was reverted

discodavey (commented):
This is just a question so I understand. Why is the grid in the article section in molecules as the content journal grid is in organisms?

viktoria-mahmud:
> This is just a question so I understand. Why is the grid in the article section in molecules as the content journal grid is in organisms?

Good point, I created a new acticle page so it is better to have new file with the styles too. You can find them in aside-section-wrapper.scss under organisms folder.

PR #1199, opened 2022-11-08
Add more generic names for classes and variables

viktoria-mahmud:


Cosmin-Daniel (approved):

source/_patterns/01-molecules/navigation/view-selector.mustache
```diff
@@ -1,13 +1,13 @@
-<div class="view-selector{{#figureUrl}} view-selector--has-figures{{/figureUrl}}" data-behaviour="ViewSelector"{{#sideBySideUrl}} data-side-by-side-link="{{sideBySideUrl}}"{{/sideBySideUrl}}>
+<div class="view-selector{{#secondaryUrl}} view-selector--has-secondary{{/secondaryUrl}}" data-behaviour="ViewSelector"{{#sideBySideUrl}} data-side-by-side-link="{{sideBySideUrl}}"{{/sideBySideUrl}}>
   <ul class="view-selector__list">
-    <li class="view-selector__list-item view-selector__list-item--article{{^figureIsActive}} view-selector__list-item--active{{/figureIsActive}}">
-      <a href="{{articleUrl}}" class="view-selector__link view-selector__link--article"><span>Article</span></a>
+    <li class="view-selector__list-item view-selector__list-item--main{{^secondaryIsActive}} view-selector__list-item--active{{/secondaryIsActive}}">
+      <a href="{{mainUrl}}" class="view-selector__link view-selector__link--main"><span>Article</span></a>
```
nlisgo, line 4: If we intend to use the view-selector template do we need to also add a parameter for `mainUrlLabel` and `secondaryUrlLabel`?

source/_patterns/01-molecules/navigation/view-selector.yaml
```diff
@@ -4,13 +4,13 @@ properties:
   sideBySideUrl:
     type: string
     minLength: 1
-  articleUrl:
+  mainUrl:
```
nlisgo, line 7: I think perhaps this should be `primaryUrl`.

source/_patterns/01-molecules/navigation/view-selector.mustache
viktoria-mahmud: Thanks, it is part of the [PR](https://github.com/elifesciences/pattern-library/pull/1178) for Tabs in Home page but it is better to be part of this one

viktoria-mahmud: Done, hardcoded text is removed

source/_patterns/01-molecules/navigation/view-selector.yaml
viktoria-mahmud: All variables that contain "main" are changed to "primary"

source/_patterns/01-molecules/navigation/view-selector.yaml
```diff
@@ -4,15 +4,21 @@ properties:
...
     minLength: 1
-  figureUrl:
+  secondaryUrl:
     type: string
     minLength: 1
-  figureIsActive:
+  secondaryIsActive:
     type: boolean
     default: false
+  primaryViewSelector:
```
nlisgo, line 16: This would be better as `PrimaryLabel` and will also need to be required.

viktoria-mahmud: Done, primaryViewSelector is renamed to primaryLabel and it is required, secondaryViewSelector is renamed to secondaryLabel

source/_patterns/01-molecules/navigation/view-selector.json
```diff
@@ -1,6 +1,6 @@
 {
-  "articleUrl": "#article",
-  "figureUrl": "#figures",
+  "primaryUrl": "#article",
+  "secondaryUrl": "#figures",
```
nlisgo, line 3: Need to add `primaryLabel` and `secondaryLabel` to this default json example.

nlisgo (approved):

nlisgo:
@viktoria-mahmud if you can just verify that you are ok with the few recent changes I've made. Happy to discuss any of them. If you are ok we can merge in the morning.

@Cosmin-Daniel I can offer some support on the patterns-php integration.

viktoria-mahmud:
> @viktoria-mahmud if you can just verify that you are ok with the few recent changes I've made. Happy to discuss any of them. If you are ok we can merge in the morning.

I am happy with these changes, thank you!

nlisgo:
Because we are changing an existing template we should probably get a head start on the patterns-php and journal PR's before we merge this in. If we merge this in but there is a delay on the patterns-php and journal then we may block other work getting into the journal.

PR #1202, opened 2022-11-17
Rename Content header classes

viktoria-mahmud:


Cosmin-Daniel (approved):

source/_patterns/02-organisms/content-headers/content-header-journal.mustache
```diff
@@ -89,33 +89,33 @@
       </div>
     </div>
     {{^isOneColumn}}
-      <div class="content-header-grid__side">
+      <div class="grid-side">
```
discodavey, line 92: Can we name this as side-section

viktoria-mahmud: Thanks, "side-section" sounds better, I rename all classes to start like this

PR #1203, opened 2022-11-22
Change style for Share modal window for all screen sizes

viktoria-mahmud:
Change Share modal style


Cosmin-Daniel (approved):

assets/sass/patterns/atoms/modal-window.scss
```diff
@@ -27,19 +27,38 @@
...
       grid-row: 4;
       justify-content: center;
+
+      @media only all and (min-width: #{get-rem-from-px($bkpt-site--medium)}em) {
+        grid-row: unset;
+        justify-content: left;
+
+        li {
+          @include margin(4, "left");
+          @include margin(4, "right");
```
Cosmin-Daniel, line 51: We can have this on a single row.

viktoria-mahmud: Done

Cosmin-Daniel (approved):

chris-huggins:
Hi @viktoria-mahmud and @Cosmin-Daniel. The new generic modal component is looking good. The body text in the generic component content doesn't appear to have any styling/correct typeface. But I assume with with it being the generic component that the content will sort that out? If not please take a look at the body text styling.


On the share modal, on larger screens there's quite a large space between the text box and button (see below). I believe this is due to the grid used to position the content. They should be nearer to provide more visual relationship between the two components.

image


This could be resolved by setting `.grid-template-columns` if it hasn't already to either a fixed size for the button (e.g. shown below using 1fr 224px). Or it might be better to use something to make the right column fit to the button and use margin on either the text box or button to get to 36px space between the two (e.g. 1fr auto might work?) 

image


Thanks.

viktoria-mahmud:
> Hi @viktoria-mahmud and @Cosmin-Daniel. The new generic modal component is looking good. The body text in the generic component content doesn't appear to have any styling/correct typeface. But I assume with with it being the generic component that the content will sort that out? If not please take a look at the body text styling.
> 
> On the share modal, on larger screens there's quite a large space between the text box and button (see below). I believe this is due to the grid used to position the content. They should be nearer to provide more visual relationship between the two components.
> 
> image
> 
> This could be resolved by setting `.grid-template-columns` if it hasn't already to either a fixed size for the button (e.g. shown below using 1fr 224px). Or it might be better to use something to make the right column fit to the button and use margin on either the text box or button to get to 36px space between the two (e.g. 1fr auto might work?)
> 
> image
> 
> Thanks.

That is a really good advice, Chris, it is not a problem to change it but it can impact the Citations modal window (they share "modal-content" and "modal-content__body" and I will update them both). I will push my changes but if you are not happy with them, I will revert them.

chris-huggins (approved):

assets/js/components/Modal.js
```diff
@@ -37,7 +37,7 @@ module.exports = class Modal {
   }
 
   modalToggle(event) {
-    if (!this.$elm.classList.contains('small-device') || this.window.innerWidth < 480) {
+    if (!this.$elm.classList.contains('small-device')) {
```
nlisgo, line 40: How will removing the additional condition affect the behaviour.

assets/sass/patterns/atoms/modal-window.scss
```diff
@@ -87,18 +106,35 @@
 }
 
 .modal-content__body {
+  @include margin(30, "top");
```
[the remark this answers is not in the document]
viktoria-mahmud: We need to re-style also the Cite variant and it has the same margin, according to the new design: https://projects.invisionapp.com/d/main#/console/22177282/470591563/preview

assets/js/components/Modal.js
viktoria-mahmud: It is a mistake, we need to keep the old logic for the share modal to be visible only on small screens but if we remove the class "small-device", we can reuse the logic for the new button and it will be visible on all screen sizes, the condition is reverted, there is no point of removing it

nlisgo (approved):

nlisgo:
Just a minor adjustment. Happy for this to be merged.

PR #1207, opened 2022-12-06
Add button collection variant with action buttons

viktoria-mahmud:


QueenKraken:
I've reviewed and I am happy.

source/_patterns/01-molecules/components/button-collection.mustache
```diff
@@ -1,5 +1,5 @@
-<ol class="button-collection{{#centered}} button-collection--centered{{/centered}}{{#compact}} button-collection--compact{{/compact}}{{#classes}} {{classes}}{{/classes}}">
-  {{#buttons}}
+<ol class="button-collection{{#variant}} button-collection--{{variant}}{{/variant}}{{#classes}} {{classes}}{{/classes}}">
```
nlisgo, line 1: The downside of this approach is that it only allows you to set one variant. Should we be allowing multiple variants?

viktoria-mahmud: good point, I can see that "button-collection--centered" and "button-collection--compact" can be used together so it is better to saparate the logic

nlisgo (approved):

PR #1208, opened 2022-12-07
Add logic for Contextual data with comma separator

viktoria-mahmud:


Cosmin-Daniel (approved):
Looks good to me.

source/_patterns/01-molecules/components/contextual-data.mustache
```diff
@@ -1,7 +1,7 @@
 <div class="contextual-data">
 
   {{#metricsData}}
-    <ul class="contextual-data__list" aria-label="The following contains the number of views, citations and annotations in this article">
+    <ul class="contextual-data__list{{#commaSeparator}} comma-separator{{/commaSeparator}}" aria-label="The following contains the number of views, citations and annotations in this article">
```
nlisgo, line 4: Just checked with @chris-huggins and @discodavey. Would have brought you in on that chat but think it's too late for you.

We should pair a little on this tomorrow. I'm not sure if we need this added parameter. I think the best option if to just introduce the template to aside and the style the metrics differently if it is found within the aside component.

Can discuss tomorrow.

discodavey: As discussed yesterday, make the CSS happen so if the parent is the aside then you can change it. Do not think there is any need to add an extra class.

PR #1209, opened 2022-12-08
Add collapsible functionality

Cosmin-Daniel:


assets/sass/patterns/atoms/definition-list.scss
```diff
@@ -17,6 +77,10 @@
   dt {
     @include font-size-and-vertical-height(14, 24);
     @include margin(0, "top");
+
+    &:first-of-type {
```
viktoria-mahmud, line 81: Do we need that truncate-with-ellipsis for the last element, when the text is too long?

assets/js/components/Collapsible.js
```diff
@@ -0,0 +1,145 @@
...
+      this.listElements.length > this.maxVisibleCollapsedElementsOnMediumViewport) {
+      this.$maxVisibleCollapsedElements = this.maxVisibleCollapsedElementsOnMediumViewport;
+      this.initialise();
+    }
+
+    this.window.addEventListener('resize', this.initialiseOnResize.bind(this));
+  }
+
+  initialise() {
+    this.$elm.classList.add('collapsible--js');
```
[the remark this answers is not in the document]
Cosmin-Daniel: You're right. I've just removed that line.

chris-huggins:
Hi @Cosmin-Daniel This is looking good, assuming I should be looking at atoms > definition-list-timeline in pattern library? 

One issue I've seen when I expand the toggle (on Chrome) and then change the size of the browser to be above 900px something odd happens below the timeline. I think it's the text "Hide history" but it's cropped a little, and has an arrow over the top of it. See the bottom of the screenshot below. Otherwise I think it's working nicely.

image

Cosmin-Daniel:
> Hi @Cosmin-Daniel This is looking good, assuming I should be looking at atoms > definition-list-timeline in pattern library?
> 
> One issue I've seen when I expand the toggle (on Chrome) and then change the size of the browser to be above 900px something odd happens below the timeline. I think it's the text "Hide history" but it's cropped a little, and has an arrow over the top of it. See the bottom of the screenshot below. Otherwise I think it's working nicely.
> 
> image

Hi @chris-huggins You're right. I've fixed the issue when content expanded on wide viewport. Please have a look again.

chris-huggins (approved):

discodavey:
All looks ok to me. Perhaps @nlisgo could review the JavaScript as well?

Had a discussion with @chris-huggins this morning about the positioning of the hide button as it jumps and it's more of a accessibility issue.

We will wait until it is in the page as to whether move it back up to where the history button is. It maybe a case of we will do something different on smaller screen sizes.

viktoria-mahmud (approved):
Looks good to me

source/_patterns/00-atoms/components/definition-list.mustache
```diff
@@ -1,8 +1,14 @@
-<dl class="definition-list{{#variant}} definition-list--{{variant}}{{/variant}}">
-  {{#items}}
-    <dt>{{{term}}}</dt>{{!
-    }}{{#descriptors}}{{!
-      }}<dd>{{{.}}}</dd>{{!
-    }}{{/descriptors}}
-  {{/items}}
-</dl>
+{{#isCollapsible}}
+  <div class="definition-list__wrapper">
```
nlisgo, line 2: Do we need to introduce a wrapper of could be achieve the same behaviour with a class being added to the element with js. If you introduce the wrapper if property `isCollapsible` is true then you need to wrap a lot of styles in `.js` but this could be avoided by introducing a class with js.

assets/js/components/Collapsible.js
```diff
@@ -0,0 +1,182 @@
...
+      this.listElements.length > this.maxVisibleCollapsedElementsOnMediumViewport) {
+      this.$maxVisibleCollapsedElements = this.maxVisibleCollapsedElementsOnMediumViewport;
+      this.initialise();
+    }
+
+    this.window.addEventListener('resize', this.initialiseOnResize.bind(this));
+  }
+
+  initialise() {
+    this.$toggle = this.createToggle();
```
nlisgo, line 33: `initialise()` can be run multiple times and some repeated steps should be avoided. If we have created a toggle element before then we have no need to recreate it we can just have a way of resetting toggle. 

What is the expected behaviour on resize? If we have expanded the definition list prior to resize do we expect for it to remain expanded on resize? cc @chris-huggins

assets/js/components/Collapsible.js
```diff
@@ -0,0 +1,182 @@
...
+module.exports = class Collapsible {
+
+  constructor($elm, _window = window, doc = document) {
+
+    this.$elm = $elm;
+    this.window = _window;
+    this.doc = doc;
+
+    this.viewportWidthMedium = 729;
+    this.viewportWidthLarge = 899;
```
nlisgo, line 14: It is my understanding that we need the 2 view port sizes to amend the text on narrower screens still. Can we push back on this to simplify the js?

@chris-huggins to avoid the additional logic can we just agree on the same text for narrow and medium?

nlisgo:
After a discussion with @Cosmin-Daniel it has been decided that too much needs to be done with this to make it useful for more than this single use case. In which case we should close this PR and then introduce the javascript when the article aside is created which is under review. `ArticleAside.js` could contain the js for the timeline as well as any of the other js behaviour needed for the article aside.

In the discussion we identified a number of ways to simplify the js and rely more on css to react to the different viewport sizes. We can perhaps pair on this.

PR #1211, opened 2022-12-16
Add new Content header journal variant

viktoria-mahmud:


Cosmin-Daniel (approved):
Reviewed and looks fine.

chris-huggins:
Hi @viktoria-mahmud This is looking great, thanks. I have three minor comments, but am not sure if they relate to other PRs or are still to do.

1. Between 729px and 731px the change in layout is strange. At 730px the institutions are hidden but the larger font and larger author list remain. I imagine this is due to different in max/min media queries where some things happen above 730px and others below. Ideally at exactly 730px it would look the same as it currently does at either 729px or 731px. 

2. Vertical position of CC and OA icons below 480px is a little off. Above 480px the alignment with the DOI looks good. But below 480px the icons appear to move a few px above the DOI baseline. I'm afraid I can't see anything in the CSS that could be the cause so I'm guessing it's to do with background position of the icons.

480px:
image

479px:
image

3. Finally I wonder if the divider at the bottom needs more padding above it? Generally I've been putting 48px in designs but perhaps better to see it in context on the page first. We'll be introducing tabs below the content header to it may need removing later anyway.

I don't mind if you want to merge this and address these issues later. Thanks.

viktoria-mahmud:
> Hi @viktoria-mahmud This is looking great, thanks. I have three minor comments, but am not sure if they relate to other PRs or are still to do.
> 
> 1. Between 729px and 731px the change in layout is strange. At 730px the institutions are hidden but the larger font and larger author list remain. I imagine this is due to different in max/min media queries where some things happen above 730px and others below. Ideally at exactly 730px it would look the same as it currently does at either 729px or 731px.
> 2. Vertical position of CC and OA icons below 480px is a little off. Above 480px the alignment with the DOI looks good. But below 480px the icons appear to move a few px above the DOI baseline. I'm afraid I can't see anything in the CSS that could be the cause so I'm guessing it's to do with background position of the icons.
> 
> 480px: image
> 
> 479px: image
> 
> 3. Finally I wonder if the divider at the bottom needs more padding above it? Generally I've been putting 48px in designs but perhaps better to see it in context on the page first. We'll be introducing tabs below the content header to it may need removing later anyway.
> 
> I don't mind if you want to merge this and address these issues later. Thanks.


Hi Chris, thank you for these comments! 

1.  It was hard for me to reproduce that behaviour because it is a corner case - you need to open the component on a smaller screen (for example 728px) and then to resize it to 730px. I can see the same behavior on the old component on prod, so it is not a new issue. Maybe it will be better to create a bug and to hande it later, what do you think?
2. Fixed.
3. I suggest to check again on the page with aside component

chris-huggins (approved):
@viktoria-mahmud Thanks. Please feel free to deploy and raise bug ticket for the other issue you mentioned.

viktoria-mahmud:
@nlisgo, I'm happy with the latest changes.

source/_patterns/02-organisms/content-headers/content-header-journal~has-aside.json
```diff
@@ -0,0 +1,348 @@
+{
+  "title": "<i>Homo naledi</i>, a new species of the genus <i>Homo</i> from the Dinaledi Chamber, South Africa",
+  "longTitle": true,
```
nlisgo, line 3: I don't think this is necessary but we don't need to clean it up here.

nlisgo (approved):

PR #1210, opened 2022-12-13
Add aside component

viktoria-mahmud:

PR #1212, opened 2022-12-21
Add JS to content-aside and functionality for timeline component on mobile

Cosmin-Daniel:


nlisgo:
I will postpone reviewing until after the following is merged: https://github.com/elifesciences/pattern-library/pull/1210

Also, could do with @chris-huggins taking a look.

chris-huggins (commented):
Please see comments on https://github.com/elifesciences/issues/issues/7935

assets/sass/patterns/organisms/content-aside.scss
```diff
@@ -29,15 +29,14 @@
   }
 
   .contextual-data__list {
```
discodavey, line 31: Can we not have this CSS in the contextual data css file rather than having it in two places

discodavey (commented):
When you reduce browser size the toggle does not kick in

discodavey (commented):
When on breakpoints 730-900 the toggle only hides one of the items where as below 730 it hides all. 

I think this is making things too complex. Either this toggle should show all or hide all.

May need to speak to @chris-huggins on this as this maybe something he may have wanted.

assets/sass/patterns/organisms/content-aside.scss
viktoria-mahmud: I'm not sure I understand correctly - do we need to move just the "contextual-data__list" css to the existing file "contextual-data.scss" ?

Cosmin-Daniel:
> When on breakpoints 730-900 the toggle only hides one of the items where as below 730 it hides all.
> 
> I think this is making things too complex. Either this toggle should show all or hide all.
> 
> May need to speak to @chris-huggins on this as this maybe something he may have wanted.

Yes, these were the requests. You can check the design [here](https://projects.invisionapp.com/d/main#/console/22177282/470910740/preview).

assets/sass/patterns/organisms/content-aside.scss
viktoria-mahmud: Done, contextual-data css is removed from the content-aside.scss file to the contextual-data.scss file

discodavey (commented):
Toggle is broken below 320. I know we support from 320 up but can we make the toggle work show/hide below 320 as well.

discodavey:
I'm happy with the changes.

discodavey (approved):

nlisgo:
If you start on the desktop view and resize the browser window to the narrow viewport the expectation is that the toggle would be there and the default would show one item in timeline for narrow and three for medium. Actual behaviour is that the toggle doesn't appear in that scenario. The toggle only appears if browser window is not wide on page load.

chris-huggins:
@nlisgo Yes they work as expected and look great, thank you.

nlisgo:
@Cosmin-Daniel can you verify that you are happy with the current state of this and approve if so.

Cosmin-Daniel:
> @Cosmin-Daniel can you verify that you are happy with the current state of this and approve if so.

Everything seems to work ok. There is only the situation when the last DT is on more than one row, but if you've talked with Chris and he agreed for this behaviour then I'm ok with it.

nlisgo:
@chris-huggins just added one more commit to remove the truncate with ellipsis on the first dt on smaller viewport sizes. I understand from our conversation yesterday that that needed to go.

PR #1218, opened 2023-01-13
Change Related item component

viktoria-mahmud:


nlisgo:
Apologies that I'm only getting around to viewing this but this is actually not needed in my opinion. The whole of the markup we use for the related item on article pages is actually using an existing component.

https://github.com/elifesciences/journal/blob/develop/src/Controller/ArticlesController.php#L168

The mustache template this populates is: https://github.com/elifesciences/pattern-library/blob/develop/source/_patterns/02-organisms/components/listing-teasers.mustache

We could perhaps repurpose this to example of a listing-teasers component that appears like the related item currently does. But the particular variant of the teaser is available here: https://ui-patterns.elifesciences.org/patterns/01-molecules-teasers-teaser-35-related-item/01-molecules-teasers-teaser-35-related-item.html

nlisgo:
Let's discuss tomorrow.

assets/sass/layout/_grid.scss
```diff
@@ -87,11 +87,11 @@
 }
 
 .grid-secondary-column__item--wide-only {
+  display: none;
 
-  @media only all and (max-width: #{get-rem-from-px($bkpt-site--wide - 1)}em) {
-    display: none;
+  @media only screen and (min-width: #{get-rem-from-px($bkpt-site--wide)}em) {
+    display: block;
```
nlisgo, line 93: This is the only change that needs scrutiny really. What is the reason for this as this class is used in journal.

See: https://github.com/elifesciences/journal/blob/986eb9b8197f8dbe0a17e0ef0ea5877dce70c0f5/app/Resources/views/article-text.html.twig#L49

Can you explain why this change is needed and what effect it will have on journal?

All other changes. My only suggestion is that we could also add an example of a related item with a see more link as this is quite common on the journal article pages.

If @chris-huggins is happy that every is as expected on the affected pages then this can be merged.

viktoria-mahmud: > This is the only change that needs scrutiny really. What is the reason for this as this class is used in journal.

The class "grid-secondary-column__item--wide-only" is used only with Related items and contains style to hide the content until 899px and to show it above 900px. When we are using max-width to apply for all screens sizes until 900px - 1, converting to rems doesn't work for 899px, I tried to avoid it but I also can see there is a lot of max-width using. If it's not an issue, I can revert it.

viktoria-mahmud: I created new example only for research article pages with related item and see more link.

nlisgo (approved):

Cosmin-Daniel (approved):

PR #1219, opened 2023-01-17
Fix Highlights on Safari

Cosmin-Daniel:


viktoria-mahmud (approved):
I can't check it on Safari but it still looks fine on other browsers

chris-huggins:
Thanks @Cosmin-Daniel. It appears to have been solved, but there's something pushing down the title of the third link and I can't tell why, or if it's related. It does appear in other browsers though, so I assume not related to the bug? But seems to be a result of this PR.

image

Cosmin-Daniel:
> Thanks @Cosmin-Daniel. It appears to have been solved, but there's something pushing down the title of the third link and I can't tell why, or if it's related. It does appear in other browsers though, so I assume not related to the bug? But seems to be a result of this PR.
> 
> image

@chris-huggins you are right, the content wasn't well aligned. I've fixed that. Could you please check it again?

chris-huggins (approved):

chris-huggins:
Thanks @Cosmin-Daniel Looks good to me too.

discodavey:
Looking in pattern library in Organisms the full highlights (three items) and also two items are doing weird things as well.
Screenshot 2023-01-18 at 11 13 00
Screenshot 2023-01-18 at 11 16 23

discodavey (commented):
All looks ok with me

PR #1220, opened 2023-01-18
Adjust JS logic to support two ways to trigger the Download popup

viktoria-mahmud:


assets/js/components/ArticleDownloadLinksList.js
```diff
@@ -14,12 +14,16 @@ module.exports = class ArticleDownloadLinksList {
     this.doc = doc;
     this.$elm = $elm;
 
-    this.$sideSectionWrapper = this.doc.querySelector('.side-section-wrapper__list');
-    if (!this.$sideSectionWrapper) {
+    this.$buttonWrapper = this.doc.querySelector('.side-section-wrapper__list') ||
+        this.doc.querySelector('.button-collection--inline');
```
Cosmin-Daniel, line 18: I think we can use the id here rather than add a new class. @viktoria-mahmud @nlisgo what do you think?

assets/js/components/ArticleDownloadLinksList.js
```diff
@@ -14,12 +14,16 @@ module.exports = class ArticleDownloadLinksList {
...
+
+    if (!this.$buttonWrapper) {
       return;
     }
 
-    this.$downloadLink = this.$sideSectionWrapper.querySelector('.side-section-wrapper__download_link');
+    this.$downloadLink = this.$buttonWrapper.querySelector('.side-section-wrapper__download_link') ||
+        this.$buttonWrapper.querySelector('.button-download');
```
Cosmin-Daniel, line 25: I think we can use the id here rather than add a new class. @viktoria-mahmud @nlisgo what do you think?

viktoria-mahmud: It's good idea to use the id, that will reduce changed files and we won't need additonal work for Pattern PHP

Cosmin-Daniel (approved):

nlisgo (approved):

PR #1228, opened 2023-02-01
Add breadcrumb for Content header

viktoria-mahmud:


smoqadam (approved):

nlisgo:
This looks good. As far as I can tell the adjustments to mustache templates are just introducing new optional parameter or making a required parameter optional. Without a change to patterns-php this should work the same on journal.

@viktoria-mahmud are you happy to pair with @smoqadam on the patterns-php piece?

viktoria-mahmud:
> This looks good. As far as I can tell the adjustments to mustache templates are just introducing new optional parameter or making a required parameter optional. Without a change to patterns-php this should work the same on journal.
> 
> @viktoria-mahmud are you happy to pair with @smoqadam on the patterns-php piece?

Sure, we can pair when @smoqadam is available next week.

PR #1233, opened 2023-02-20
Add jump link

viktoria-mahmud:


assets/sass/patterns/molecules/jump-menu.scss
```diff
@@ -0,0 +1,54 @@
...
+
+  &:hover {
+    color: $color-text;
+  }
+}
+
+.jump-menu__active {
+  @include margin(-2, "left");
+  color: $color-text;
+  border-left: 2px solid $color-primary;
```
discodavey, line 32: Needs to be in alpha order

discodavey (commented):
There is some spacing still that needs removing on article-research mustache file

discodavey:
@QueenKraken This is part of some bigger work with putting tabs into the journal like we have in EPP. Is this still a priority to do after the decision was made last week to scale back Journal work?

discodavey:
I know that we are using the view selector code for the jump links but is everyone happy with this approach or would we consider using the existing code for both i.e. could this be done or is there a better way of doing this?

This is purely a throwing this up in the air. There is no right or wrong answer to this

assets/sass/patterns/molecules/jump-menu.scss
viktoria-mahmud: Done.

viktoria-mahmud:
Done.

> There is some spacing still that needs removing on article-research mustache file

viktoria-mahmud:
We need to support View Selector but it will be removed later, so it will be easier to remove the logic. 

> I know that we are using the view selector code for the jump links but is everyone happy with this approach or would we consider using the existing code for both i.e. could this be done or is there a better way of doing this?
> 
> This is purely a throwing this up in the air. There is no right or wrong answer to this

discodavey (approved):

PR #1237, opened 2023-02-23
Adjust Related item margin

viktoria-mahmud:


chris-huggins (approved):

source/_patterns/02-organisms/components/content-aside~complete.json
```diff
@@ -110,6 +110,10 @@
           }
         }
       }
-    ]
+    ],
+    "seeMoreLink": {
+      "name": "Further reading",
+      "url": "#"
```
E-Kemp, line 116: Where will this URL come from?

source/_patterns/04-pages/article--research~with-content-aside-with-scroll.json
```diff
@@ -311,7 +311,11 @@
             }
           }
         }
-      ]
+      ],
+      "seeMoreLink": {
+        "name": "Further reading",
+        "url": "#"
```
E-Kemp, line 317: Ditto to previously, I don't know where this URL comes from, maybe I'm missing some key understanding?

source/_patterns/02-organisms/components/content-aside~complete.json
viktoria-mahmud: Often in Pattern library we are not using real links.

E-Kemp:
LGTM

assets/sass/patterns/atoms/see-more-link.scss
```diff
@@ -10,6 +10,9 @@
     content: "\00a0\00bb";
   }
 
+  .content-aside__related-item & {
+    @include margin(24, "top");
```
discodavey, line 14: Can we not add this to the bottom of the  .listing-list__item class  rather than putting the margin on the top?

We try and put the margin and padding on the bottom of items unless there is no other way

discodavey (changes requested):
One other thing. There is a div on it's own without any class or id. Please can we remove this.
Screenshot 2023-02-28 at 11 54 10

assets/sass/patterns/atoms/see-more-link.scss
viktoria-mahmud: I added the margin on the top because "see-more-link" is optional and it will increase the distance after the component. It is not a problem to decrease component bottom space and add it on "listing-list__item" class.

viktoria-mahmud:
> One other thing. There is a div on it's own without any class or id. Please can we remove this. Screenshot 2023-02-28 at 11 54 10


There is a logic for that div if Highlights exists. It requires more investigation to fix that issue.

image

discodavey:
> > One other thing. There is a div on it's own without any class or id. Please can we remove this. Screenshot 2023-02-28 at 11 54 10
> 
> There is a logic for that div if Highlights exists. It requires more investigation to fix that issue.
> 
> image

Can you not wrap the whole div in the highlights tag and the same for the closing tag as well?

viktoria-mahmud:
> Can you not wrap the whole div in the highlights tag and the same for the closing tag as well?

Sure, I didn't think of that option, thanks!

assets/sass/patterns/atoms/see-more-link.scss
viktoria-mahmud: Done.

discodavey (approved):

PR #1239, opened 2023-02-24
Add different font size for Status title

viktoria-mahmud:


nlisgo:
There are 3 possibilities with the change to the schema.

- no titleLength
- `long` titleLength
- `short` titleLength

Will each of these be styled differently or is `short` treated the same as omitting the property?

viktoria-mahmud:
> There are 3 possibilities with the change to the schema.
> 
> * no titleLength
> * `long` titleLength
> * `short` titleLength
> 
> Will each of these be styled differently or is `short` treated the same as omitting the property?

Yes, there is no difference if the we have short title lenght or no title lenght.

source/_patterns/02-organisms/components/content-aside~status-title-long.json
```diff
@@ -0,0 +1,48 @@
+{
+  "status": {
+    "title": "Author Accepted Manuscript",
```
nlisgo, line 3: The logic in patterns-php should set the titleLength to long if above 23 characters.

nlisgo (approved):

PR #1240, opened 2023-03-01
Author popup fixes

viktoria-mahmud:


assets/js/components/Popup.js
```diff
@@ -222,13 +222,31 @@ module.exports = class Popup {
...
 
+  hasAuthorClassName() {
+    const authorLink = document.querySelector('.author_link');
+    if (!authorLink) {
+      return false;
+    }
+
+    const authorLinkParent = authorLink.parentElement.parentElement;
+    const hasAuthorItem = authorLinkParent.classList.contains('author_list_item');
+    return hasAuthorItem;
```
nlisgo, line 233: Why are we simply not returning true if `.author_link` is found in the document?

nlisgo: It feels a bit wrong to be making this change because Popup.js is supposed to be a generic behaviour that can be applied to any component and not need to have override's for specific cases.

We should add examples in for the now 3 article page templates:

- new with content aside
- feature article
- insight or editorial

We should try and keep all of the override code together and we should make it clear that it is because of the position absolute.

nlisgo:
The removal of `position: relative` only appeared to affect content headers with a background image. As no article page has a background image in the content header, could this work?

https://github.com/elifesciences/pattern-library/pull/1243

viktoria-mahmud:
The solution was introduced on different [PR](https://github.com/elifesciences/pattern-library/pull/1243).

PR #1242, opened 2023-03-02
Alternative approach to refactoring ContentAside.js

nlisgo:


viktoria-mahmud:
The logic is working as expected, it looks good to me.

PR #1244, opened 2023-03-07
Add aria-label attribute

viktoria-mahmud:


discodavey (commented):
Looks good to me. 

Just wondering if we put "Descriptive and accessible name for the link" then I think it gives it purpose and the developer knows why it's there for.

discodavey (approved):

discodavey (approved):

PR #1245, opened 2023-03-07
Add aria-label for Timeline links

viktoria-mahmud:


discodavey (commented):
Once again I would put descriptive and accessible name.

discodavey (commented):
Last one... I would put Descriptive and accessible name

PR #1246, opened 2023-03-08
Feature/accessibility teaser links

viktoria-mahmud:


source/_patterns/01-molecules/teasers/teaser.mustache
```diff
@@ -5,7 +5,7 @@
   {{#contextLabel}}
     <ol class="teaser__context_label_list" aria-label="These research categories are for the following article">
       {{#list}}
-        <li class="teaser__context_label_item">{{#url}}<a href="{{url}}" {{#carouselItem}}tabindex="-1"{{/carouselItem}} class="teaser__context_label_link">{{/url}}{{{name}}}{{#url}}</a>{{/url}}</li>
+        <li class="teaser__context_label_item">{{#url}}<a href="{{url}}" {{#contextAriaLabel}}aria-label="{{contextAriaLabel}}" {{/contextAriaLabel}}{{#carouselItem}}tabindex="-1" {{/carouselItem}}class="teaser__context_label_link">{{/url}}{{{name}}}{{#url}}</a>{{/url}}</li>
```
discodavey, line 8: Is there any reason why we have added tabindex -1?

discodavey (commented):
All good apart from the connversation needed about the tabindex being -1. This means that the user will not be able to get to this link which I personally think is bad.

I know this is legacy code.

discodavey (approved):
Approved but will need to investigate tabindex in carouselItem ticket as to whether it should stay or be removed.

PR #1252, opened 2023-03-20
Add tab link component

viktoria-mahmud:


discodavey (commented):
Can we please have the same naming convention as in EPP.

assets/sass/patterns/molecules/tab-menu.scss
```diff
@@ -0,0 +1,68 @@
...
+  @media only all and (min-width: #{get-rem-from-px($bkpt-site--extended-wide)}em) {
+    grid-column: 1/9;
+  }
+
+  @media only all and (min-width: #{get-rem-from-px($bkpt-site--x-wide)}em) {
+    grid-column: 1/10;
+  }
+
+  & ~ .view-selector,
+  & ~ .main-content-grid {
```
discodavey, line 55: Have we over complicated using these? Do we need them?

source/_patterns/04-pages/article--research.mustache
```diff
@@ -26,6 +26,10 @@
 
            <div class="content-container-grid">
 
+             {{#tabMenu}}
+               {{> molecules-tab-menu }}
+             {{/tabMenu}}
```
discodavey, line 31: These are not showing in article research

discodavey (commented):
Tabs are not level on acrtive with the not active tabs. Text jumps up slightly. Looks like padding.

Also moves to the right as well. Take a look at the epp version to reference

discodavey (commented):
Grey borders need to be full width until 730px

discodavey:
There spacing between the blue bar and border. This should be removed

Screenshot 2023-03-28 at 09 09 20

discodavey:
Should the example in molecule functional. As it working if you click each of the tabs?

discodavey:
```secondary-column-grid``` is showing but is not doing anything. Is this related to the ticket?

File: assets/sass/patterns/molecules/tab-menu.scss (hunk 1 above)
Comment (discodavey) [reply assets/sass/patterns/molecules/tab-menu.scss:55]: Can we not have view-selector and main-content-grid as grid-row: 2?

File: source/_patterns/02-organisms/components/content-aside.mustache [hunk 3]
```diff
@@ -3,15 +3,15 @@
     <div class="content-aside__first-column">
       {{#status}}
         <div class="status-wrapper">
-          <h5 class="status-title{{#titleLength}} status-title__{{titleLength}}{{/titleLength}}">{{status.title}}</h5>
+          <h5 class="status-title{{#titleLength}} status-title__{{titleLength}}{{/titleLength}}">{{title}}</h5>
```
discodavey, line 6: Is this anything to do with the tab menu build?

viktoria-mahmud:
@chris-huggins, can you advise me on this case - on [EPP ](https://elifesciences.org/reviewed-preprints/84918/reviews) it is full width but on the [design ](https://projects.invisionapp.com/d/main#/console/22152008/470052610/inspect)it is not, which one is the correct one?

> Grey borders need to be full width until 730px

source/_patterns/02-organisms/components/content-aside.mustache
viktoria-mahmud: No, it is not related.

assets/sass/patterns/molecules/tab-menu.scss
viktoria-mahmud: If we don't have Tabs, this style won't apply and will be back to grid-row: 1. It is basically ".tab-menu__nav ~ .main-content-grid"

viktoria-mahmud:
I can see the tab links redirect to another page on EPP, so we can test this on Journal, on Pattern library we have only default "#" value and the logic is not working.

> Should the example in molecule functional. As it working if you click each of the tabs?

viktoria-mahmud:
It is fixed, now the only difference beween active and non active link is the bold text.

> Tabs are not level on acrtive with the not active tabs. Text jumps up slightly. Looks like padding.
> 
> Also moves to the right as well. Take a look at the epp version to reference
> 
> Screenshot 2023-03-28 at 09 01 26

discodavey (commented):
The active tab text does not line up with the non active text.

Screenshot 2023-03-29 at 07 54 33

viktoria-mahmud:
They are aligned, the only difference is the bold text for an active link. I can add some paddings but that will add jump effect.

image



> The active tab text does not line up with the non active text.
> 
> Screenshot 2023-03-29 at 07 54 33

discodavey:
@viktoria-mahmud If you look at the EPP version you will see what I mean.

assets/js/components/TabbedNavigation.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+module.exports = class ContentAside {
```
discodavey, line 3: Could the class name be better?

viktoria-mahmud: Fixed.

assets/sass/patterns/molecules/article-section.scss
```diff
@@ -48,6 +48,10 @@ $arrow-size: 5;
   // Edge case: prevent space above the h2s of the first header  in the first article section
   & .article-section__header:first-child h2 {
     @include nospace("top");
+
+    .wrapper--content-with-header-and-aside & {
```
discodavey, line 52: Is this for something else apart from the tabs?

discodavey (commented):
Make the top and bottom borders between 480-730px full width

viktoria-mahmud:
Tab component is on full width between 480-730px (up to730px in total as it is on EPP).

> Make the top and bottom borders between 480-730px full width

assets/sass/patterns/molecules/article-section.scss
viktoria-mahmud: Yes, it is releted to article pages with aside component.

discodavey:
I think that is everything. If @chris-huggins is happy then I think we are good.

discodavey (approved):

assets/js/components/TabbedNavigation.js
```diff
@@ -0,0 +1,28 @@
...
+
+    const lastActiveElement = this.$elm.querySelector('.tabbed-navigation__tab-label--active a');
+
+    if (!lastActiveElement || e.target === lastActiveElement) {
+      return;
+    }
+
+    lastActiveElement.parentNode.classList.remove(this.activeClassName);
+    e.target.parentNode.classList.add(this.activeClassName);
+  }
```
nlisgo, line 27: I'm not clear how this is supposed to work in production. Where is the javascript to perform the important operation of retrieving and displaying the content? What is the plan to get this ready for production?

nlisgo (changes requested):

assets/sass/patterns/organisms/content-header-journal.scss
```diff
@@ -326,10 +326,6 @@
 
     .divider {
       display: none;
-
-      @media only all and (min-width: #{get-rem-from-px($bkpt-site--extended-wide)}em) {
-        display: block;
-      }
```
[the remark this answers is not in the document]
viktoria-mahmud: The style for divider will apply only in "content-header__has-aside" class, which is used only in pages with Content header and Aside component, so it won't affect other pages.

We don't need to hide/show the sections below, tab click should trigger page reload, as it is on [EPP](https://elifesciences.org/reviewed-preprints/84918).
@chris-huggins, please, correct me if I'm wrong.

assets/js/components/TabbedNavigation.js
viktoria-mahmud: I think these tabs should do the same as View selector primary and secondary links - redirect to another page or "jump" to section with specific ID (for example click on a link with href=#articleUrl, will navigate to a section with the same ID).

nlisgo: > I think these tabs should do the same as View selector primary and secondary links - redirect to another page or "jump" to section with specific ID (for example click on a link with href=#articleUrl, will navigate to a section with the same ID).

@chris-huggins would this be acceptable?

assets/sass/patterns/organisms/content-header-journal.scss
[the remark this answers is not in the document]
nlisgo: You seem to confirm these are breaking changes. This is probably an example of a change that we need to see working in journal before we merge it into pattern-library. It may not affect other pages but it will affect article pages with a content aside unless they have this new component. This component is not ready. The js disables the behaviour of the links so they will not be followed. This was presumably to demonstrate the intended behaviour but it prevents it from following the links.

assets/js/components/TabbedNavigation.js
chris-huggins: @nlisgo Yes I think that makes sense. 

i.e. Figures and data should go to the separate page in the same way as the current selector 

We would need a new separate page for "Peer Review" similar to on EPP. We would need to move the content from the lower Peer Review and Author Response article sections to this new tab/page.

A final tab would switch to side-by-side view, which I guess is also a new page (but wouldn't have tabs introduced). 


I would like to avoid a tab causing scrolling down the page.

chris-huggins (approved):

nlisgo (approved):

PR #1255, opened 2023-03-21
Support big image for all Research article pages

viktoria-mahmud:


discodavey (approved):
Let's get Chris to review this via the demo link and then we can push

chris-huggins:
@discodavey Looks ok to me on pattern library at least

PR #1257, opened 2023-03-21
Change button text to See less

viktoria-mahmud:


discodavey:
I do not think just the arrow on it's own is enough. Does not look obvious and you could miss it if you scan the page which most users do. @chris-huggins do you think we need "see more" on 320 to 729?

Screenshot 2023-03-22 at 08 30 54

discodavey:
@chris-huggins do you think a little more spacing is needed on the left to mirror the spacing on the right?

Screenshot 2023-03-22 at 08 43 23

chris-huggins:
@discodavey Yes. Although I had imagined keeping it below the names as it was before. I don't mind if we stick with it after the last name with some additional space, but if the button has to wrap onto a new line then I'd like to avoid a space preceding it.


i.e. If we do this (where red block shows space after last name):
image


I worry we'll end up with space in front of the button when it ends up on a new line. Which I'd like to avoid:
image

I think the solution would be to center it always on a new line. 
image

discodavey:
@viktoria-mahmud Can you look at what @chris-huggins has put above and chat to him if you have any questions please.

chris-huggins:
@viktoria-mahmud Sorry if this wasn't clear. I only meant to change it below 730px where we currently have the issue with the small arrow button. I'm now seeing it centered above 730px too, but was only looking to change it below 730px.

So to clarify, could you please return it to the 'normal' position we had before (following last author) above 730px. 

I'd also like to see the arrows continue to point up below 730px please.

If it's easier to change to an SVG let me know.

discodavey:
Arrow does not expand when clicked on.

Screenshot 2023-03-23 at 13 58 41

viktoria-mahmud:
@chris-huggins Is it expected behavior? It is old logic but I can fix it.

> Arrow does not expand when clicked on.
> 
> Screenshot 2023-03-23 at 13 58 41

chris-huggins:
@viktoria-mahmud I'm happy to stick with just `et al.` as the link to show the additional names below 730px. At the moment the arrows are not visible below 730px when they are collapsed. I think it's just on this PR that they are visible so could be easier to hide them?

To clarify:

Below 730px default - no need for arrows (screenshot is production but has too big a gap):

image


Below 730px 'opened' - arrows should be visible and pointing up.
image

viktoria-mahmud:
@chris-huggins thanks, the arrow is removed. Also the padding before and after the link for all screen sizes is reduced from 6px to 3px.

> @viktoria-mahmud I'm happy to stick with just `et al.` as the link to show the additional names below 730px. At the moment the arrows are not visible below 730px when they are collapsed. I think it's just on this PR that they are visible so could be easier to hide them?
> 
> To clarify:
> 
> Below 730px default - no need for arrows (screenshot is production but has too big a gap):
> 
> image
> 
> Below 730px 'opened' - arrows should be visible and pointing up. image

discodavey (approved):

discodavey:
@viktoria-mahmud Has Chris approved this now?

I have tested the removal of the two lines of code and all still works.

PR #1282, opened 2023-06-16
Correct downstream build job name in Jenkinsfile.prod

nlisgo:


nlisgo:
@lsh-0 I'm not sure if this was caused by recent updates but we have this seems to fix the failing builds in ci.

See: https://alfred.elifesciences.org/job/prod-pattern-library/

lsh-0:
This was probably me, moving archived/dead jobs into [the `/archived` folder](https://alfred.elifesciences.org/job/Archived/), investigating broken builds and I'm pretty sure I moved some top-level jobs with names prefixed with `dependencies-` or `release-` into the `/dependencies` and `/release` folders.

I can't see any other references to this specific job.

PR #1294, opened 2023-08-15
Change DOIs to Version DOI

smoqadam:


discodavey (approved):
Looks good from what I have tested on

PR #1304, opened 2024-02-21
Adjust z-index so tabbed navigation does not show when search overlay…

discodavey:
… appears


chris-huggins:
@discodavey Looks good to me, thanks!

viktoria-mahmud (approved):
Looks good.

PR #1305, opened 2024-02-22
fix: Due to cropping issue

discodavey:


discodavey:
@chris-huggins Does this look ok?

https://s3.amazonaws.com/ci-pattern-library/1305/patterns/04-pages-home/04-pages-home.html

If so then I will push it through to journal for a final check

chris-huggins:
@discodavey Looks good to me, thanks.

PR #1314, opened 2024-05-29
Timeline with version and color

viktoria-mahmud:


assets/sass/patterns/atoms/definition-list.scss
```diff
@@ -3,34 +3,71 @@
...
+      &:after {
+        @include padding(0 5 0);
+        content: "\00a0\2219\00a0";
+        color: $color-text-secondary;
+        font-weight: bold;
+      }
+    }
+
     a {
+      @include padding(3, "top");
```
discodavey, line 38: Can we remove the padding from the top and add it to the bottom of the <time> element

assets/sass/patterns/atoms/definition-list.scss
```diff
@@ -3,34 +3,71 @@
...
+    @include padding(28, "left");
     font-family: $font-secondary;
     font-weight: bold;
+    position: relative;
   }
 
   dt {
-    @include font-size-and-vertical-height(14, 24);
+    @include font-size-and-vertical-height(16, 24);
     @include margin(18, "top");
```
discodavey, line 14: Can we make this so the margin is bottom and not top.

assets/sass/patterns/organisms/content-aside.scss
```diff
@@ -124,72 +132,56 @@
...
-        }
-      }
     }
 
     .toggle {
-      @include font-size-and-vertical-height(12, 24);
-      @include padding(0 17 0 10);
-      align-self: start;
+      @include font-size-and-vertical-height(14, 24);
+      @include margin(18, "top");
```
discodavey, line 165: Can we put the margin top on the bottom of the dl?

viktoria-mahmud: Discussed, we will keep the margin top because the toggle is optional and won't be visible when there is one element in the timeline.

assets/sass/patterns/atoms/definition-list.scss
viktoria-mahmud: Discussed, we will keep the margin top because the toggle is optional and won't be visible when there is one element in the timeline.

assets/sass/patterns/atoms/definition-list.scss
viktoria-mahmud: That padding is removed as not needed.

PR #1324, opened 2024-08-21
Add Previous version warning

viktoria-mahmud:


discodavey:
The text is left aligned in the atom but should be centered until it goes into the aside column when it's left aligned

